PR #3, opened 2020-08-04
feat(mongo): support offset and limit

skjindal93:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -355,4 +367,13 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
   public void drop() {
     collection.drop();
   }
+
+  // cap the max limit to 10k records
+  private Integer normalizeLimit(Integer limit) {
+    if (limit != null && limit > MAX_LIMIT) {
```
aaron-steinfeld, line 373: Why the max limit? there previously was no max limit (since a cursor doesn't require one for perf reasons) - is there a compelling reason to add one?

skjindal93: no specific reason. I remembered we have some 10K limit in Pinot as well. Anyways, we can leave that to the consumers of Mongo

_getting rid of the MAX LIMIT_

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoDocStoreTest.java
```diff
@@ -64,6 +64,32 @@ public void cleanup() {
...
+    collection.upsert(new SingleValueKey("default", "testKey2"), createDocument("foo2", "bar2"));
+    collection.upsert(new SingleValueKey("default", "testKey3"), createDocument("foo3", "bar3"));
+    collection.upsert(new SingleValueKey("default", "testKey4"), createDocument("foo4", "bar4"));
+    collection.upsert(new SingleValueKey("default", "testKey5"), createDocument("foo5", "bar5"));
+
+    Query query = new Query();
+    query.setLimit(2);
+    query.setOffset(1);
+
+    Iterator<Document> results = collection.search(query);
```
kotharironak, line 80: Can you extend this test to the same query twice with pagination? That will cover the default order test.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -157,7 +156,19 @@ private String decodeKey(String key) {
...
     DBObject ref = new BasicDBObject(map);
-    final DBCursor dbCursor = collection.find(ref);
+    DBCursor cursor = collection.find(ref);
+
+    Integer offset = query.getOffset();
+    if (offset != null && offset >= 0) {
+      cursor = cursor.skip(offset);
+    }
+
+    Integer limit = query.getLimit();
```
inespo, line 166: Offset and limit should've been done AFTER orderBy. The update is done in the cursor, and not passed to the mongo query. It's better to pass the offset, limit and orderBy to Docstore query. This would perform better and you don't need to worry about the order when it's applied. You can separate the 2 issues though. Most urgent is to fix the offset and limit AFTER sorting.

document-store/src/test/java/org/hypertrace/core/documentstore/mongo/MongoCollectionTest.java
```diff
@@ -122,4 +135,87 @@ public void testParseNestedQuery() {
...
+    DBCursor cursor = mock(DBCursor.class);
+    when(collection.getFullName()).thenReturn("I am Collection");
+    when(collection.find(any())).thenReturn(cursor);
+    mongoCollection.search(query);
+
+    verify(cursor, times(0)).skip(anyInt());
+  }
+
+  @Test
+  public void testOffsetAndLimitInQuery() {
```
inespo, line 192: Please add a test to use offset, limit and orderBy in 1 query.

PR #4, opened 2020-08-10
Count documents based on query

skjindal93:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -321,6 +321,19 @@ public long count() {
     return collection.count();
   }
 
+  @Override
+  public long count(Query query) {
```
kotharironak, line 325: Now, our quey interface has a limit and offset, how would it affect here?
1. document here that count refers to all search criteria documents
2. we pass limit/offset (skip) as options in the collection. count API via `DBCollectionCountOptions`

skjindal93: Renamed it to total to make better sense. We need to find the total number of documents, irrespective of the limit/offset applied. Hence, the total query would just read filters from the incoming query and ignore everything else

PR #5, opened 2020-08-24
chore: added utility functions for order by

skjindal93:


document-store/src/main/java/org/hypertrace/core/documentstore/Query.java
```diff
@@ -19,6 +19,14 @@ public void setFilter(Filter filter) {
     this.filter = filter;
   }
 
+  public void addAllOrderBys(List<OrderBy> orderBys) {
+    this.orderBys.addAll(orderBys);
+  }
+
+  public void addOrderBy(OrderBy orderBy) {
```
avinashkolluru, line 26: looks similar to setOrderBy below. Do you want to delete the below one?

skjindal93: yeah. let me do that

avinashkolluru (approved):

PR #7, opened 2020-08-31
chore(mongo): case insensitive like search

skjindal93:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -279,7 +279,9 @@ private void parseOrderByQuery(List<OrderBy> orderBys, Map<String, Object> order
           map.put(filter.getFieldName(), value);
           break;
         case LIKE:
-          map.put(filter.getFieldName(), new BasicDBObject("$regex", value));
+          // Case insensitive regex search
```
skjindal93, line 282: Only logic change. Rest are tests and formatting changes

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -279,7 +279,9 @@ private void parseOrderByQuery(List<OrderBy> orderBys, Map<String, Object> order
...
+          map.put(
```
kotharironak, line 283: would this leave something is searched case insensitive in a document store, and but OLAP at case sensitive?  

Does LIKE is generally case-insensitive? Or we kind need both kind of filter `LIKE` and `LIKEI`?

skjindal93: OLAP is already case insensitive

kotharironak (approved):
lgtm

PR #9, opened 2020-11-09
Adding postgres document store

jainnidhi703:
This PR addresses the issue: https://github.com/hypertrace/hypertrace/issues/114


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -0,0 +1,120 @@
...
+   * @return Returns Tables for a particular database
+   */
+  @Override
+  public Set<String> listCollections() {
+    Set<String> collections = new HashSet<>();
+    try {
+      DatabaseMetaData metaData = client.getMetaData();
+      ResultSet tables = metaData.getTables(null, null, "%", new String[]{"TABLE"});
+      while (tables.next()) {
+        collections.add(this.database + "." + tables.getString("TABLE_NAME"));
```
kotharironak, line 61: Q: will the collection interface impl take care of any table? or It will be restricted to the table described in https://github.com/hypertrace/hypertrace/issues/114 (just having JSONB column)

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -0,0 +1,120 @@
...
+import java.sql.*;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class PostgresDatastore implements Datastore {
+  
+  private static final Logger LOGGER = LoggerFactory.getLogger(PostgresDatastore.class);
+  
+  private static final String DEFAULT_DB_NAME = "default_db";
```
jcchavezs, line 19: I wonder if this is actually needed. It looks a bit arbitrary to me (unless I am missing something). In the docs I could only find the `postgres` as the only default DB: https://www.postgresql.org/docs/9.1/app-initdb.html

jcchavezs: I see this is to keep parity with https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoDatastore.java#L18 but do we want to repeat that, is there any true use case where there is a `default_db` somewhere ready to be used? cc @adriancole @aaron-steinfeld

aaron-steinfeld: I don't have the context on why this was there (just made my first commit to doc store this week!) but I'd agree with what I think you're saying - we shouldn't make any assumptions about the database structure in the code, and if for some reason we need to, it better match the default installation of the implementation database. It looks like here we take in an alternative via config, which is better than the mongo implementation.

jainnidhi703: This was just kept for parity in reference to MongoDB, Postgres supports default connection, and I agree with you on that. Will make the change.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
jainnidhi703: The collection name and column type(json/jsonb) are configurable, but this collection impl will only support the above schema. If there is anything missing in the above schema, we can discuss that. Also, open to making it configurable in any way possible.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
jcchavezs: I am sorry I did not explain myself correctly when I pointed out postgres documentation. My thoughts were that we should not make any assumption on users defaults and that the only default database I heard about was this `postgres` DB. I would suggest that we do not use any default database name and instead we require it from user as even if postgres provides one, user might decide to delete it. Instead if we ask the user to pass one database name, they should take care of that.

jainnidhi703: Currenlty, I have added the functionality where user can pass the database from config. In case the user doesn't pass any database name, then it fallbacks to Postgres's default database i.e. `postgres`.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+      preparedStatement.setString(3, jsonString);
+      ResultSet resultSet = preparedStatement.executeQuery();
+      
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Write result: " + resultSet.toString());
+      }
+      
+      return resultSet.rowUpdated();
+    } catch (SQLException e) {
+      LOGGER.error("SQLException inserting document. key: {} content:{}", key, document, e);
```
kotharironak, line 58: There are few places, it seems that the interface `upsert` is not handled for return type `boolean` and relying on exception - e.g https://github.com/hypertrace/entity-service/blob/main/entity-service-impl/src/main/java/org/hypertrace/entity/type/service/EntityTypeServiceImpl.java

So, do you see any set of the exceptions are error state and can be thrown back as IOException? 
Or we can catch all and rethrow? But would it raise an exception on updating the same value if an entry already exists?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/postgres/PostgresDocStoreTest.java
```diff
@@ -0,0 +1,159 @@
+package org.hypertrace.core.documentstore.postgres;
```
kotharironak, line 1: Just a note: For this test to work, we will need a similar task or extension via Postgres docker image in `build.gradle.kts`. we can do that at the end once everything is fine. For now, we can test by starting the server locally.

e.g in build.gradle.kts for mongo
```
tasks.register<DockerPullImage>("pullMongoImage") {
  image.set("mongo:4.2.0")
}
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
jcchavezs: As I said we don't necessarily have to repeat what we did in the past related to this, I'd rather not have a default value and require it from user cc @kotharironak @buchi-busireddy

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+      .stream()
+      .map(orderBy -> orderBy.getField() + " " + (orderBy.isAsc() ? "ASC" : "DESC"))
+      .filter(str -> !str.isEmpty())
+      .collect(Collectors.joining(" , "));
+    
+    return orderBySQL;
+  }
+  
+  @Override
+  public boolean delete(Key key) {
```
kotharironak, line 226: Can we add one test for delete as well?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/postgres/PostgresDocStoreTest.java
```diff
@@ -0,0 +1,159 @@
...
+    postgresConfig.putIfAbsent("user", "postgres");
+    postgresConfig.putIfAbsent("password", "postgres");
+    postgresConfig.putIfAbsent("type", "jsonb");
+    Config config = ConfigFactory.parseMap(postgresConfig);
+    
+    datastore = DatastoreProvider.getDatastore("Postgres", config);
+    System.out.println(datastore.listCollections());
+  }
+  
+  @AfterEach
```
kotharironak, line 44:
```suggestion
  @BeforeEach
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/postgres/PostgresDocStoreTest.java
```diff
@@ -0,0 +1,159 @@
...
+  public void cleanup() {
```
kotharironak, line 45:
```suggestion
  public void setUp() {
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+  public long count() {
+    String countSQL = String.format("SELECT COUNT(*) FROM %s", collectionName);
+    try {
+      PreparedStatement preparedStatement = client.prepareStatement(countSQL);
+      ResultSet resultSet = preparedStatement.executeQuery();
+      return resultSet.getLong(1);
+    } catch (SQLException e) {
+      LOGGER.error("SQLException deleting all documents.", e);
+    }
+    return 0;
```
kotharironak, line 267: Interesting!! One more case, I guess need to correct the contract.
Q: Can we go with `-1` as an error case?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+  }
+  
+  @Override
+  public long total(Query query) {
+    StringBuilder totalSQLBuilder = new StringBuilder("SELECT COUNT(*) FROM ").append(collectionName);
+    long count = 0;
```
kotharironak, line 273:
```suggestion
    long count = -1;
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+    
+    return null;
+  }
+  
+  @VisibleForTesting
+  /**
+   *
+   */
+  String parseQuery(Filter filter) {
+    if (filter.isComposite()) {
```
jcchavezs, line 121: I think this can break down into smaller static methods to improve readability something like this::parseQueryForCompositeFilter and this::parseQueryForNonCompositeFilter

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: We have to create an issue in hypertrace to handle the return value correctly. Will you create one?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+    return false;
+  }
+  
+  @Override
+  public boolean updateSubDoc(Key key, String subDocPath, Document subDocument) {
+    return false;
+  }
+  
+  @Override
+  public Iterator<Document> search(Query query) {
```
kotharironak, line 69: Currently, there is an assumption in HT code that each document will have an _id field.
e.g https://github.com/hypertrace/entity-service/blob/572b6eaa4db0b898967258faa7f3f12f43d122cc/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java#L533 
we can see that the usage of constant `EntityServiceConstants.ID` in the above method. It is added to the query filter. We need to create one more issue in HT to fix this.

But, for now, I would suggest if we can add the _id field in each document in the `upsert` and `bulkupsert` method.
```
String jsonString = document.toJson();
// convert to JSON, and we can add root level field in document
{
_id : key.toString()
// rest of document document.toJson()
}
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+          break;
+        case NOT_EXISTS:
+          // TODO: Checks if key does not exist
+          break;
+        case AND:
+        case NEQ:
+        case OR:
+          throw new UnsupportedOperationException("Only Equality predicate is supported");
+        default:
+          break;
```
kotharironak, line 196: In the default case, should we return the `null` string?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -0,0 +1,347 @@
...
+          }
+        }
+        case AND: {
+          String childList =
+            Arrays.stream(filter.getChildFilters())
+              .map(this::parseQuery)
+              .filter(str -> !str.isEmpty())
+              .collect(Collectors.joining(" AND "));
+          if (!childList.isEmpty()) {
+            return childList;
```
kotharironak, line 143: Q: Do we need bracketing here as child filter can also be composite? - `"("+ childList + ")"`
So, case like e.g `(a=2 AND b=3) or (c=4 AND d=5)`, is this fine as `a=2 AND b=3 OR c=4 AND d=5?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/postgres/PostgresDocStoreTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.io.IOException;
+import java.util.*;
+
+public class PostgresDocStoreTest {
+  public static final String ID_KEY = "id";
+  public static final String DOCUMENT_KEY = "document";
+  private static final String COLLECTION_NAME = "mytest2";
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final String UPDATED_AT = "updated_at";
+  private static final String CREATED_AT = "created_at";
```
kotharironak, line 23: It seems there are assumptions on this name too - https://github.com/hypertrace/entity-service/blob/3d1f9afc4def73d9489156af16e18920e2eb17a7/entity-service-impl/src/main/java/org/hypertrace/entity/service/constants/EntityServiceConstants.java 

To me, it looks fine. we can see if there is any issue while doing e2e test. if, so, we will change its value.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
jainnidhi703: Have removed the default value for Postgres, should I add an exception in case the user doesn't provide database name?

adriancole: drive-by 2p is that in both cases, the default DB should be overridable and "hypertrace" as it is inevitable that people will want to re-use an existing DB and not necessarily overlap. Plus when you do integration tests, it is helpful to be accustomed to never using a default db or something confusingly named default db which isn't default.

adriancole: ex in zipkin in integration tests we frequently vary the DB, of course it depends on each test gets a new container or not, but at least in logs it is still helpful sometimes.

adriancole (commented):
sometime when this goes to merge it should be considered if document-store should in the same source tree implement mongo and postgresql. The gradle project could for example, be re-organized so that the main interface and integration tests are in a common area, which more easily ensures behavior parity even if the two impls (and associated docker) are in different directories/modules. This could further allow conditonal integration test runs (ex not running mongo tests when only postgres changes. These points are mainly for consideration, and probably while this is being figured out a bit early.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Yes, we can pass null on default case.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Yes, added this.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Actually, for default, we can also throw an unsupported exception, similar to what we do for composite filters.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Have fixed this, all Exceptions will go throw and exception. Also, it won't throw an error, if it's updating the same value.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -0,0 +1,135 @@
...
+      // Database needs to be created before initializing connection
+      if (config.hasPath("database")) {
+        this.database = config.getString("database");
+        url = url + database;
+      }
+
+      if (config.hasPath("user") && config.hasPath("password")) {
+        client = DriverManager.getConnection(url, config.getString("user"), config.getString("password"));
+      } else {
+        client = DriverManager.getConnection(url, "postgres", "postgres");
```
kotharironak, line 48: @jainnidhi703 I have added this while testing locally, will check with you in the morning, and correct it as default.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -0,0 +1,135 @@
...
+  public boolean init(Config config) {
+    try {
+      DriverManager.registerDriver(new org.postgresql.Driver());
+      String url;
+      if (config.hasPath("url")) {
+        url = config.getString("url");
+      } else {
+        String hostName = config.getString("host");
+        int port = config.getInt("port");
+        url = String.format("jdbc:postgresql://%s:%s/", hostName, port);
```
kotharironak, line 36: @jainnidhi703 I have added this one as well to read based on host/port properties.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
jainnidhi703: @kotharironak should we connect without a user password, in case it is not sent from the config?

kotharironak:
@jainnidhi703 and I have gone through the final code review on HT slack channel and addressed most of the comments as part of the final commit. The final changes look good to me.

kotharironak:
@jainnidhi703 and I have gone through the final code review on HT slack channel and addressed most of the comments as part of the final commit. The final changes look good to me.

kotharironak:
> sometime when this goes to merge it should be considered if document-store should in the same source tree implement mongo and postgresql. The gradle project could for example, be re-organized so that the main interface and integration tests are in a common area, which more easily ensures behavior parity even if the two impls (and associated docker) are in different directories/modules. This could further allow conditonal integration test runs (ex not running mongo tests when only postgres changes. These points are mainly for consideration, and probably while this is being figured out a bit early.

@jainnidhi703 Can you open an issue for this and address this in follow up PR? For now, we can go forward with the existing impl to unblock deployment at your end.

kotharironak:
Verified the PR after final change e2e by manually using with docker-compose setup by doing  below steps
- pushed the artifacts to local maven
- modified config-bootstraper, attribute-service, and entity-service to consume the newer artifact
- updated the config for the above services to use postgres
- added new containers for postgres in docker-compose setup

Output:
Screenshot 2020-11-19 at 6 49 12 PM
Screenshot 2020-11-19 at 6 33 00 PM

kotharironak (approved):
lgtm (tested manually e2e with docker-compose setup)

kotharironak:
Locally ran the snyk-scan and updated postgres library to fix it.

kotharironak (approved):

jainnidhi703:
> sometime when this goes to merge it should be considered if document-store should in the same source tree implement mongo and postgresql. The gradle project could for example, be re-organized so that the main interface and integration tests are in a common area, which more easily ensures behavior parity even if the two impls (and associated docker) are in different directories/modules. This could further allow conditonal integration test runs (ex not running mongo tests when only postgres changes. These points are mainly for consideration, and probably while this is being figured out a bit early.

@adriancole Have created a separate [issue](https://github.com/hypertrace/hypertrace/issues/123#issue-747161222) for the above. Will make the changes in a separate PR.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.Set;
+import org.hypertrace.core.documentstore.Collection;
+import org.hypertrace.core.documentstore.Datastore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Provides {@link Datastore} implementation for Postgres DB.
+ */
+public class PostgresDatastore implements Datastore {
```
jcchavezs, line 26: Since this is an implementation that I believe it isn't supposed to be extended, does making it final makes sense?

kotharironak: @jainnidhi703 Can you handle this in your follow up PR?

PR #11, opened 2020-11-17
fix: set created time for bulk upsert documents

skjindal93:


buchi-busireddy (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -340,7 +340,10 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
             .upsert()
             .update(
                 new BasicDBObject("$set", dbObject)
-                    .append("$currentDate", new BasicDBObject("_lastUpdateTime", true)));
+                    .append("$currentDate", new BasicDBObject("_lastUpdateTime", true))
```
kotharironak, line 343: @skjindal93 what was the issue? It seems there are some assumptions on the code with _lastUpdateTime construct.

skjindal93: @kotharironak It isn't changing the `_lastUpdateTime` construct. It's adding the `createdTime` while bulkUpserting the documents

The `createdTime` field is already added when we `upsert` a document in Mongo https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java#L90, but the `createdTime` field was not present when the documents were bulk upserted

kotharironak: @skjindal93 But, are we using it in any filtering operations or any other query operations?

skjindal93: @kotharironak I don't think it's been used in any query operations yet.

But, since the `createdTime` field is generated when we `upsert`, it should also be generated during `bulkUpsert` for correctness

PR #15, opened 2020-12-04
Makes lastUpdatedTime consistent with createdTime

inespo:
_lastUpdateTime stores the last updated timestamp in Date and has different
convention than createdTime.

field name without underscore is preferred, and long value makes it
easier for comparison with the current type support

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -51,6 +51,7 @@ public class MongoCollection implements Collection {
   private static final Logger LOGGER = LoggerFactory.getLogger(MongoCollection.class);
   public static final String ID_KEY = "_id";
   private static final String LAST_UPDATE_TIME = "_lastUpdateTime";
+  private static final String LAST_UPDATED_TIME = "lastUpdatedTime";
   /* follow json/protobuf convention to make it deser, let's not make our life harder */
   private static final String CREATED_TIME = "createdTime";
   private static final ObjectMapper MAPPER = new ObjectMapper();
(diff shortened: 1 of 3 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -131,6 +132,7 @@ private BasicDBObject prepareUpsert(Key key, Document document) throws JsonProce
     JsonNode sanitizedJsonNode = recursiveClone(jsonNode, this::encodeKey);
     BasicDBObject setObject = BasicDBObject.parse(MAPPER.writeValueAsString(sanitizedJsonNode));
     setObject.put(ID_KEY, key.toString());
+    setObject.put(LAST_UPDATED_TIME, System.currentTimeMillis());
```
kotharironak, line 135: Can we add java doc on top of public function `upsert` that it adds two special keys on top of input document _ID and `lastUpdatedTime`?

kotharironak (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
inespo: ok, will add

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -131,11 +139,16 @@ private BasicDBObject prepareUpsert(Key key, Document document) throws JsonProce
...
```
avinashkolluru, line 142: Please call System.currentTimeMillis once and use it both the places. Based on the current code, there is a possibility that lastUpdatedTime and createTime for newly created documents will mismatch.

Also, if we plan to use the lastUpdatedTime, do we still need the older lastUpdateTime field?

document-store/src/test/java/org/hypertrace/core/documentstore/FilterTest.java
```diff
@@ -61,4 +61,10 @@ public void testSettingChildFiltersField() {
     Assertions.assertNotNull(filter2.getChildFilters());
     Assertions.assertEquals(0, filter2.getChildFilters().length);
   }
+
+  @Test
+  public void testFilterDate() {
+    Filter filter = new Filter();
```
avinashkolluru, line 67: Was this checked in by mistake?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
inespo: we will need to keep the older _lastUpdateTime for now because copying the time to lastUpdatedTime from _lastUpdateTime is a manual process. We need to give time to allow for this migration before deleting the older field.

document-store/src/test/java/org/hypertrace/core/documentstore/FilterTest.java
inespo: yes, will remove.

kotharironak (approved):

Issue #14, opened 2020-11-27
PostgresCollection implementation is vulnerable to SQLi


avinashkolluru:
See the method parseQueryForNonCompositeFilter here - https://github.com/hypertrace/document-store/blob/f9a3b6882e239b0d05b000e9db6744b268510adb/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java#L207

We seem to be appending strings to create a SQL statement instead of creating a parameterized query using `?` and later setting the corresponding arguments while querying.

This is prone to SQL injection.

JBAhire:
This has been addressed by #30 so closing this issue.

PR #16, opened 2020-12-08
Fixing SQLi Vulnerability

jainnidhi703:
As mentioned in https://github.com/hypertrace/document-store/issues/14, for the existing Postgres Implementation, while building where clause, we were appending strings that are prone to SQL Injections.
In this PR, we are building a parameterized where clause. While building the String query, we will add `?` instead of the actual value. Also, in parallel, we will build a `Map<Integer, Object>` which will consist of the index and the actual param value.
Both of the above will be used while building the `PreparedStatement`.


jainnidhi703:
https://github.com/hypertrace/document-store/issues/14

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/Params.java
```diff
@@ -0,0 +1,151 @@
+package org.hypertrace.core.documentstore.postgres;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Holds the params that need to be set in the PreparedStatement for constructing the final PQL
```
avinashkolluru, line 7: Please fix the comment. Looks like this has been copy/pasted from query-service

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -246,22 +251,23 @@ protected String parseQueryForNonCompositeFilter(Filter filter) {
...
         // TODO: Checks if key does not exist
       case NEQ:
-        throw new UnsupportedOperationException("Only Equality predicate is supported");
       default:
         throw new UnsupportedOperationException(
             String.format("Query operation:%s not supported", op));
     }
-    return filterString.append("'").append(value).append("'").toString();
+    String filters = filterString.append(QUESTION_MARK).toString();
+    paramsBuilder.addStringParam(String.valueOf(value));
```
avinashkolluru, line 259: Is every param treated as a string? Would that work?
If that is the case, why do we need the other methods like `addLongParam`/`addFloatParam` in Params?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -270,18 +276,32 @@ protected String parseQueryForCompositeFilter(Filter filter) {
...
     }
   }
 
+  @VisibleForTesting
+  protected PreparedStatement buildPreparedStatement(String sqlQuery, Params params) throws SQLException {
+    PreparedStatement preparedStatement = client.prepareStatement(sqlQuery);
+    params.getStringParams().forEach((k, v) -> {
+      try {
+        // Postgres Index starts from 1
+        preparedStatement.setString(k+1, v);
```
avinashkolluru, line 297: Since the Params class is specific to the Postgres implementation, should we initialize the default in Params to 1 and directly use k here instead of doing a `k+1` here?

avinashkolluru, line 297: Since the preparedStatement allows setting different types, we should probably call the appropriate setLong/setFloat etc methods instead of assuming every param to be a String.
I think that will be functionally correct.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/Params.java
jainnidhi703: Yes, will fix it

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Currently all params are treated, as string, have added LongParam for future use cases

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Yes, agree on that

jainnidhi703: The only approach I can think of, when we do [Object value = filter.getValue()](https://github.com/hypertrace/document-store/blob/f9a3b6882e239b0d05b000e9db6744b268510adb/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java#L209), can check the datatypes `getClass`, and then consider setLong/setFloat. Please suggest if there are any alternatives?

avinashkolluru: Yes. For now i think we should do that, until we have first class support for different types in the document-store filteer

jcchavezs:
great job @jainnidhi703 could you please describe in this PR how is this issue being fixed?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: @avinashkolluru, tried it with `getClass()`, but the issue is querying a field in jsonb document returns a text always, so I have to cast the text to respective type(float/long/int).  Example: `SELECT * FROM mytest WHERE document->>'integer' = 123`, this doesn't work directly. Need to cast i.e. `SELECT * FROM mytest WHERE (document->>'integer')::INT = 123`. And. casting is an expensive operation in postgres, do we really want to add it?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: What was the value here - `filterString.append("'").append(value).append("'").toString()`?
Here value is an `Object` so, it `toString` method used, right?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/Params.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Map;
+
+/**
+ * Holds the params that need to be set in the PreparedStatement for constructing the final SQL
+ * query
+ */
+public class Params {
+
+    // Map of index to the corresponding param value
+    private final Map<Integer, Integer> integerParams;
```
kotharironak, line 13: Currently, for the `Filter` interface, we have value as `Object` type - https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/Filter.java
We don't have knowledge of proper types and with which operator what are valid types.

So, as a thought (for avoiding type conversations w/o proper interface) should we go with just an object map here for all params and convert to its string representation during buildPreparedStatement?
```
public class Params {
private final Map<Integer, Object> objectParams;
}
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -270,18 +276,32 @@ protected String parseQueryForCompositeFilter(Filter filter) {
...
         throw new UnsupportedOperationException(
-            String.format("Boolean operation:%s not supported", op));
+            String.format("Query operation:%s not supported", op));
     }
   }
 
+  @VisibleForTesting
+  protected PreparedStatement buildPreparedStatement(String sqlQuery, Params params) throws SQLException {
+    PreparedStatement preparedStatement = client.prepareStatement(sqlQuery);
+    params.getStringParams().forEach((k, v) -> {
```
kotharironak, line 294: Based on the suggestion above on using `Map<Integer, Object>` for param, we can do as below.
```
protected PreparedStatement buildPreparedStatement(String sqlQuery, Params params) throws SQLException {
    PreparedStatement preparedStatement = client.prepareStatement(sqlQuery);
    params.getObjectParams().forEach((k, v) -> {
      try {
        // Postgres Index starts from 1
        isValidTypes(value) // true if type is string, int, long, double, false for composite type
        preparedStatement.setString(k, String.valueOf(value));
      } catch (SQLException e) {
        LOGGER.error("SQLException querying documents. query: {}", sqlQuery, e);
      }
    });
    return preparedStatement;
  }
```
A drawback of this approach is that for `>` operator, we might have a `string` value. And, some document-store might have support for this as well. So, we can just check valid types for now - string, int, long, double and discard composite Object types. What do you guys think?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -236,7 +238,10 @@ protected String parseQueryForNonCompositeFilter(Filter filter) {
         List<Object> values = (List<Object>) value;
         String collect = values
             .stream()
-            .map(val -> "'" + val + "'")
+            .map(val -> {
+              paramsBuilder.addStringParam(String.valueOf(val));
```
kotharironak, line 242: Where are we adding int/long/double params?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
avinashkolluru: If we want to check basic types, why even change it to a generic Map<Integer, Object>?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: How do we know the type of document field type? Here, we are trying to cast based on the `input` value which can be of any type as it is provided by client. We can cast if we know the destination type or it was indicated by the interface. I think this will lead to many failure cases.

In the above e.g if the amount contains both integer and float values, would casting work if the input object is of type INT?
e.g 
```
doc1
{
'amount' : 10
}

doc2 {
'amount' : 10.5
}

Filter -> Op = Eq and Value is INT object with value 10.
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: if we go with the existing definition, we will have to add casting logic while filter parsing. The `INT` value of the filter, goes into Map<Interger, Integer>. But, I don't see how we can use this specific information while doing a query. We still have to convert all of them into finally `String` as we don't know the type of destination field in `jsonb` document. So, suggesting that we can just simplify logic by replacing all Where clause filter with `?` and collecting all values into Map<Integer, Object>.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: Also, can you verify for `>` operator for `document->>'amount'` having all int values if you use string conversation? Are you getting the expected results?
e.g 
```
doc1->amount:10, doc2->amount:5, doc3->amount:7, doc4->amount:10
select * from mytest where document->>'amount' > '4' and document->>'amount' < '8'
```
Output: `doc2, doc3`

jainnidhi703: Verified the above with string, and works fine.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
jainnidhi703: Only for the cases where we might be querying `id` directly

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: As discussed above is a string comparison-based search. it won't work correctly on negative values as its sorting order is based on the string type.
e.g 
```
doc1->amount:-10, doc2->amount:5, doc3->amount:-5, doc4->amount:5.2, doc5->amount:-7, doc6->amount:7 doc7->amount:6
```

Also in the document store, the same field can have different types. Mongo handles it well, and it takes care during query time without throwing any exceptions. 
As an example, insert the below docs in your local Mongo
```
db.test.insertOne(
   { "name": "a", "qty": "b" }
)
db.test.insertOne(
   { "name": "b", "qty": "-5" }
)
db.test.insertOne(
   { "name": "c", "qty": "5" }
)
db.test.insertOne(
   { "name": "c", "qty": "7" }
)
```
Try queries:
```
db.test.find( {$and: [{ qty: { $gte: 'b' } }, { qty: { $lte: 'b' } }] }) - seaching all docs in collections where qty is string
db.test.find( {$and: [{ qty: { $gte: -5 } }, { qty: { $lte: 10 } }] }) - seaching all docs in collections where qty is numeric
```
So, even if we assume that there is no mixed type ingested in HT for any jsonb document for a given field, we still
need to handle numeric/boolean type queries for operator `EQ, GT, LT, GTE, LTE`.

So, as discussed, we can add the following template for `EQ, GT, LT, GTE, LTE` while parsing.
For numeric:
```
case jsonb_typeof(<field>)
WHEN ‘number’  THEN (<field>)::numeric  > ?
end
```
For boolean:
```
case jsonb_typeof(<field>)
WHEN 'boolean'  THEN (<field>)::boolean  > ?
end
```
For string (existing implementation or below for consistency) :
```
case jsonb_typeof(<field>)
WHEN 'string'  THEN (<filed>)::string  > ?
end
```

kotharironak:
closing this as taken care as part of - https://github.com/hypertrace/document-store/pull/30

PR #18, opened 2020-12-15
Adding a new returnAndBulkUpsert API in the doc store.

buchi-busireddy:
This will help the clients to build more interesting use cases and also cutdown a roundtrip to the server.

A use case where this has even become essential is more like a change data capture kind of use case: A client wants to upsert a bunch of docs but it needs to know what all changed in those docs compared to the previous version of the docs so that it can optimize some processing. CDC (Change data capture) at the doc store level might be helpful but that's much more heavy weight and doesn't have enough use cases currently to bring that in. 

Also, it's a known that the new API is sync API and could take a bit longer than other APIs so it should be used wisely based on the use cases.

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -412,15 +412,17 @@ public class MongoCollection implements Collection {
   }
 
   @Override
-  public Iterator<Document> bulkUpsertAndReturn(Map<Key, Document> documents) throws IOException {
+  public Iterator<Document> returnAndBulkUpsert(Map<Key, Document> documents) throws IOException {
     try {
+      // First get all the documents for the given keys.
+      FindIterable<BasicDBObject> cursor = collection.find(selectionCriteriaForKeys(documents.keySet()));
+      final MongoCursor<BasicDBObject> mongoCursor = cursor.cursor();
+
+      // Now go ahead and do the bulk upsert.
       BulkWriteResult result = bulkUpsertImpl(documents);
       LOGGER.debug(result.toString());
 
-      FindIterable<BasicDBObject> cursor = collection.find(selectionCriteriaForKeys(documents.keySet()));
-      final MongoCursor<BasicDBObject> mongoCursor = cursor.cursor();
       return new Iterator<>() {
-
         @Override
         public boolean hasNext() {
           return mongoCursor.hasNext();
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -412,7 +412,7 @@ public class MongoCollection implements Collection {
   }
 
   @Override
-  public Iterator<Document> returnAndBulkUpsert(Map<Key, Document> documents) throws IOException {
+  public Iterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Document> documents) throws IOException {
     try {
       // First get all the documents for the given keys.
       FindIterable<BasicDBObject> cursor = collection.find(selectionCriteriaForKeys(documents.keySet()));
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -30,6 +30,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import net.jodah.failsafe.Failsafe;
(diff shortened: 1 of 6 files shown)


jcchavezs:
I am usually reluctant to such additions as it forces sync processing which
isn't good for high load. Could you please describe some of the uses cases
you have in mind?

buchi-busireddy:
> I am usually reluctant to such additions as it forces sync processing which isn't good for high load. Could you please describe some of the uses cases you have in mind?

@jcchavezs updated the description with more details.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -445,6 +429,58 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
...
+
+  @Override
+  public Iterator<Document> returnAndBulkUpsert(Map<Key, Document> documents) throws IOException {
+    try {
+      String collect = documents.keySet().stream()
+          .map(val -> "'" + val.toString() + "'")
+          .collect(Collectors.joining(", "));
+
+      String space = " ";
+      String query = new StringBuilder("SELECT * FROM")
```
tim-mwangi, line 458: This seems like a shortcut to SQL injection, no?

buchi-busireddy: Hmm.. Thanks for pointing. Let me check and think more..

buchi-busireddy: I think the SQLi is being fixed as a part of https://github.com/hypertrace/document-store/pull/16/files
I'll merge this and will update the other PR and will fix SQLi as a part of that. If the other PR is merged before mine, i can update and fix it.

tim-mwangi (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -386,28 +387,58 @@ public long total(Query query) {
...
+          prepareUpsert(key, entry.getValue()),
+          new UpdateOptions().upsert(true)));
+    }
+
+    return Failsafe.with(bulkWriteRetryPolicy)
+        .get(() -> collection.bulkWrite(bulkCollection, new BulkWriteOptions().ordered(false)));
+  }
+
+  @Override
+  public Iterator<Document> returnAndBulkUpsert(Map<Key, Document> documents) throws IOException {
```
avinashkolluru, line 415: nit: rename to bulkUpsertAndReturnOlderDocuments?
returnAndBulkUpsert sounds a bit wierd.

buchi-busireddy: Done.

avinashkolluru (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -386,28 +387,58 @@ public long total(Query query) {
...
+  @Override
+  public Iterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Document> documents) throws IOException {
+    try {
+      // First get all the documents for the given keys.
+      FindIterable<BasicDBObject> cursor = collection.find(selectionCriteriaForKeys(documents.keySet()));
+      final MongoCursor<BasicDBObject> mongoCursor = cursor.cursor();
+
+      // Now go ahead and do the bulk upsert.
+      BulkWriteResult result = bulkUpsertImpl(documents);
+      LOGGER.debug(result.toString());
```
jcchavezs, line 423: I am not sure about the performance of this. Not a Javaer here but it seams whether logger is enabled or not we still turn it into string? cc @kotharironak

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -386,28 +387,58 @@ public long total(Query query) {
...
+          return mongoCursor.hasNext();
+        }
+
+        @Override
+        public Document next() {
+          return MongoCollection.this.dbObjectToDocument(mongoCursor.next());
+        }
+      };
+    } catch (JsonProcessingException e) {
+      LOGGER.error("Error during bulk upsert for documents:{}", documents, e);
```
jcchavezs, line 437: This is more like an in general comment. I am usually in favour of either log the error and handle it or bubble up the exception but not both of them because they usually flood the logs. Also, do we want to print the full set of documents in logs? How about privacy concerns and also efficient usage of the log storage? I don't thing dumping the failing documents in the logs is actionable either.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -386,28 +387,58 @@ public long total(Query query) {
...
+      throw new IOException("Error during bulk upsert.");
```
jcchavezs, line 438: Not passing the previous exception makes us loosing all the context on this error. Is there any reason for not doing it?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -417,6 +448,11 @@ private BasicDBObject selectionCriteriaForKey(Key key) {
     return new BasicDBObject(ID_KEY, key.toString());
   }
 
+  private BasicDBObject selectionCriteriaForKeys(Set<Key> keys) {
```
jcchavezs, line 451: Can't this be private? A simple inspection tells me yes it does but I am not 100% sure. Tho `ID_KEY` is static.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -445,6 +429,58 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
...
+
+  @Override
+  public Iterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Document> documents) throws IOException {
+    try {
+      String collect = documents.keySet().stream()
+          .map(val -> "'" + val.toString() + "'")
+          .collect(Collectors.joining(", "));
+
+      String space = " ";
+      String query = new StringBuilder("SELECT * FROM")
```
jcchavezs, line 458: I usually have reluctancies with the `select *`, do we really need to pass it all, can't we make explicit the fields we pass so that changes in the schema don't affect this kind of methods silently?

jcchavezs:
@buchi-busireddy I am sorry I am late to the party but I left you some comments.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
aaron-steinfeld: 👍 we should always be letting the logger doing the stringification for us so we don't have to eat this cost unless the message is needed. That means wrapping it in an `if` or IMO, more graceful to do `LOGGER.debug("{}", result);`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
buchi-busireddy: @jcchavezs The need here is indeed the full document since we want to return all the fields of the document's current copy, as the API requires it.
At this layer, we don't know what all fields are present in the document actually.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: Fixing in a new PR. Thanks!

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: I was thinking the API should mask the implementation specific exception details but this is actually a library so I'll fix it.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: I think you meant to ask about static right? Fixing it.

jcchavezs: yeah I meant static, sorry.

jcchavezs:
Hi @buchi-busireddy I wonder if you had time to put up some of the fixed we talked about in this PR post merge.

buchi-busireddy:
@jcchavezs Thanks for reminding and sorry that I missed raising the PR. I do have the changes locally and will raise shortly.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: Raised https://github.com/hypertrace/document-store/pull/25

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: Raised https://github.com/hypertrace/document-store/pull/25

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: Raised https://github.com/hypertrace/document-store/pull/25

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
buchi-busireddy: https://github.com/hypertrace/document-store/pull/25

Issue #19, opened 2020-12-18
Provides supports for contains, exists and not_exists filter in postgres


kotharironak:
Document store is exposing the following filters for query interface - https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/Filter.java

Postgres implementation should also consider support for 'contains', 'exists' and 'not_exists' filter.
Recently, these filters are also exposed as part of entity-service - https://github.com/hypertrace/entity-service/pull/62.

suresh-prakash:
Support added in https://github.com/hypertrace/document-store/pull/151

PR #20, opened 2020-12-21
feat: support NEQ operator in mongo implementation of document-store interface

skjindal93:
- Adds support for NEQ operator for MongoDB document store implementation
- Postgres side implementation will be taken care of as part of https://github.com/hypertrace/document-store/issues/19


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -352,10 +352,15 @@ private void parseOrderByQuery(List<OrderBy> orderBys, Map<String, Object> order
         case NOT_EXISTS:
           map.put(filter.getFieldName(), new BasicDBObject("$exists", false));
           break;
-        case AND:
         case NEQ:
+          // $ne operator in Mongo also returns the results, where the key does not exist in the
```
buchi-busireddy, line 356: @skjindal93 But, by definition, the records that don't contain the key should be returned for NE right?

skjindal93: Depends on the definition of NE. I was thinking it of as a relational database use case.

To be honest, even I was in a dilemma what to do here, but then if we don't have the `exists: true` filter, the consumers would have to know, that the query will hit document store and should add the `exists: true` filter specifically

What we can do is the following though. I can add the exists filter in entity service in the NEQ operator implementation there, and keep it as just `$ne` in document store

I think that would make more sense. What do you think?

kotharironak: +1

It depends on how we want to define the NEQ operator at the document store level. But, I think, EQ and NEQ are similar in semantics. So, whatever EQ matches, the remaining items in a set should match by NEQ

And for a document store, a collection is a set of documents. However, there is no schema for a document in a collection. So, a set can have differnt documents as well.  As an e.g
```
db.test.insertOne({ "greet" : "hello" })
db.test.insertOne({ "first" : "f_one",  "last":"l_one", "qty":1})
db.test.insertOne({ "first" : "f_two",  "last":"l_two", "qty":-1})
```
And, I see that MongoDB is doing the same for a set.
```
Q: db.test.find( { qty: { $eq: 1 } } )
R:{ "first" : "f_one",  "last":"l_one", "qty":1}

Q: db.test.find( { qty: { $neq: 1 } } )
R: { "greet" : "hello" }, { "first" : "f_two",  "last":"l_two", "qty":-1}
```
So, we should also follow a similar definition if we are mapping to `eq` (EQ operator of document-store) or `ne` operator (NEQ of document-store). We can have a new operator if we need the `NEQ_EXISTS` filter for a special case scenario.

kotharironak: >What we can do is the following though. I can add the exists filter in entity service in the NEQ operator implementation > there, and keep it as just $ne in document store

At the `entity-service` level, it makes sense as it knows about a set of documents, but was thinking instead of overloading the operator at any level,  can the query be formed by add both the filters based on need using the `AND` operator - `AND (NEQ xyz) (EXISTS xyz)`?

skjindal93: > can the query be formed by adding both the filters based on need using the `AND` operator -` AND (NEQ xyz) (EXISTS xyz)`

But, doesn't that mean, the consumer knows it's querying a document store and hence had to add that extra `EXISTS` filter?

kotharironak: Yes. the consumer already knows that it is using the `document-store` interface and it also knows its own need. 
Secondly, the consumer (here entity-service) controls the data set and can have its own `NEQ` or `EQ` based on that. But, it seems that we are having certain documents (created by ES) that can have missing `keys` and it is not able to enforce a schema. In that, case this - `AND (NEQ xyz) (EXISTS xyz)`- seems to be a better option at the ES level where we will still have the flexibility to fetch all docs with `NEQ` semantics.

buchi-busireddy: +1 to what Ronak said. In this case, EDS is the consumer and it knows it's using a doc store so it should and those conditions.

kotharironak: @skjindal93 Updated the PR for addressing the above comments. Made `NEQ` opposite to `EQ`.

kotharironak (approved):

PR #24, opened 2020-12-30
chore(CI): adds googlejavaformat action to CI.

jcchavezs:
This PR adds a google java format check on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - "**/*.md"
+      - "LICENSE"
+  pull_request:
+
+jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: axel-op/googlejavaformat-action@v3
+        with:
+          args: "--set-exit-if-changed"
--- a/document-store/build.gradle.kts
+++ b/document-store/build.gradle.kts
@@ -6,99 +6,99 @@ import com.bmuschko.gradle.docker.tasks.network.DockerCreateNetwork
 import com.bmuschko.gradle.docker.tasks.network.DockerRemoveNetwork
 
 plugins {
-  `java-library`
-  jacoco
-  id("org.hypertrace.publish-plugin")
-  id("org.hypertrace.jacoco-report-plugin") version "0.1.0"
-  id("org.hypertrace.integration-test-plugin") version "0.1.0"
-  id("com.bmuschko.docker-remote-api") version "6.4.0"
+    `java-library`
+    jacoco
+    id("org.hypertrace.publish-plugin")
+    id("org.hypertrace.jacoco-report-plugin") version "0.1.0"
+    id("org.hypertrace.integration-test-plugin") version "0.1.0"
+    id("com.bmuschko.docker-remote-api") version "6.4.0"
 }
 
 dependencies {
-  api("com.typesafe:config:1.3.2")
-  implementation("org.postgresql:postgresql:42.2.13")
-  implementation("org.mongodb:mongodb-driver-sync:4.1.1")
-  implementation("com.fasterxml.jackson.core:jackson-databind:2.11.0")
-  implementation("org.slf4j:slf4j-api:1.7.25")
-  implementation("com.google.guava:guava-annotations:r03")
-  implementation("org.apache.commons:commons-lang3:3.10")
-  implementation("net.jodah:failsafe:2.4.0")
-  testImplementation("org.junit.jupiter:junit-jupiter:5.6.2")
-  testImplementation("org.mockito:mockito-core:2.19.0")
-  integrationTestImplementation("org.junit.jupiter:junit-jupiter:5.6.2")
-  integrationTestImplementation("com.github.java-json-tools:json-patch:1.13")
+    api("com.typesafe:config:1.3.2")
+    implementation("org.postgresql:postgresql:42.2.13")
+    implementation("org.mongodb:mongodb-driver-sync:4.1.1")
+    implementation("com.fasterxml.jackson.core:jackson-databind:2.11.0")
+    implementation("org.slf4j:slf4j-api:1.7.25")
+    implementation("com.google.guava:guava-annotations:r03")
+    implementation("org.apache.commons:commons-lang3:3.10")
+    implementation("net.jodah:failsafe:2.4.0")
+    testImplementation("org.junit.jupiter:junit-jupiter:5.6.2")
+    testImplementation("org.mockito:mockito-core:2.19.0")
+    integrationTestImplementation("org.junit.jupiter:junit-jupiter:5.6.2")
+    integrationTestImplementation("com.github.java-json-tools:json-patch:1.13")
 }
 
 tasks.test {
-  useJUnitPlatform()
+    useJUnitPlatform()
 }
 
 tasks.register<DockerCreateNetwork>("createIntegrationTestNetwork") {
-  networkName.set("document-store-int-test");
+    networkName.set("document-store-int-test");
 }
 
 tasks.register<DockerRemoveNetwork>("removeIntegrationTestNetwork") {
-  networkId.set("document-store-int-test")
+    networkId.set("document-store-int-test")
 }
 
 tasks.register<DockerPullImage>("pullMongoImage") {
-  image.set("mongo:4.4.0")
+    image.set("mongo:4.4.0")
 }
 
 tasks.register<DockerCreateContainer>("createMongoContainer") {
-  dependsOn("createIntegrationTestNetwork")
-  dependsOn("pullMongoImage")
-  targetImageId(tasks.getByName<DockerPullImage>("pullMongoImage").image)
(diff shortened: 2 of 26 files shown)


jcchavezs:
@aaron-steinfeld could you please loop in and tell what are the rules we expect to meet? also the version for the formatting rules?

aaron-steinfeld:
> @aaron-steinfeld @buchi-busireddy could you please loop in and tell what are the rules we expect to meet? also the version for the formatting rules?

We generally use https://plugins.jetbrains.com/plugin/8527-google-java-format - so it looks like that currently means 1.9 (I think their versioning is not of the rules themselves which don't appear to be versioned https://google.github.io/styleguide/javaguide.html , but the tool's implementation?)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
```
aaron-steinfeld, line 7: Do we need ignore paths? This check takes ~15s, and the problem with limiting actions based on changed paths is that we can't do the same with the protected branch required checks. So if this action sometimes doesn't run, we can't require it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
...
+      - "**/*.md"
+      - "LICENSE"
+  pull_request:
+
+jobs:
+  formatting:
+    runs-on: ubuntu-latest
```
aaron-steinfeld, line 14: please pin the env (we've been using `ubuntu-20.04` elsewhere)

document-store/build.gradle.kts
```diff
@@ -6,99 +6,99 @@ import com.bmuschko.gradle.docker.tasks.network.DockerCreateNetwork
 import com.bmuschko.gradle.docker.tasks.network.DockerRemoveNetwork
 
 plugins {
-  `java-library`
-  jacoco
-  id("org.hypertrace.publish-plugin")
-  id("org.hypertrace.jacoco-report-plugin") version "0.1.0"
-  id("org.hypertrace.integration-test-plugin") version "0.1.0"
-  id("com.bmuschko.docker-remote-api") version "6.4.0"
+    `java-library`
```
aaron-steinfeld, line 9: Is this to verify the check is working? The indentation should be 2 https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation

skjindal93:
Is there a way to auto format the code using some command line option? 

I recently added https://github.com/diffplug/spotless in one of the repositories, which provides a command line option to auto format the entire code and it has various formatting options, google formatting being one of them https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format

We can add git hooks and auto apply/check the formatting as `pre-push` or `pre-commit`. Though, applying formatting on staged files is still an open issue in spotless https://github.com/diffplug/spotless/issues/623, but we could still add a spotless check hook till then https://github.com/diffplug/spotless/issues/178#issuecomment-351638034

JBAhire:
> Is there a way to auto format the code using some command line option?
> 
> I recently added https://github.com/diffplug/spotless in one of the repositories, which provides a command line option to auto format the entire code and it has various formatting options, google formatting being one of them https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format
> 
> We can add git hooks and auto apply/check the formatting as `pre-push` or `pre-commit`. Though, applying formatting on staged files is still an open issue in spotless [diffplug/spotless#623](https://github.com/diffplug/spotless/issues/623), but we could still add a spotless check hook till then [diffplug/spotless#178 (comment)](https://github.com/diffplug/spotless/issues/178#issuecomment-351638034)
> 
> I think @pavolloffay also uses the same plugin in Java Agent https://github.com/hypertrace/javaagent/blob/a339bd5747af6cef4fbc666e641278b61cfbc273/gradle/spotless.gradle

@skjindal93, are you talking about something like this: https://github.com/thegreystone/action-jmc-spotless ?

Not sure if this helps. If not we can still write shellscript and use it with dockerfile and create our own action similar to above. 

I believe it will be series of bash commands so we can have basic dockerfile with gradle as in above action and then add bash script as `entrypoint.sh`. would that help? 

will be happy to work with you if this is significant and can help us in future as well. we can take a stab.

skjindal93:
> > Is there a way to auto format the code using some command line option?
> > I recently added https://github.com/diffplug/spotless in one of the repositories, which provides a command line option to auto format the entire code and it has various formatting options, google formatting being one of them https://github.com/diffplug/spotless/tree/main/plugin-gradle#google-java-format
> > We can add git hooks and auto apply/check the formatting as `pre-push` or `pre-commit`. Though, applying formatting on staged files is still an open issue in spotless [diffplug/spotless#623](https://github.com/diffplug/spotless/issues/623), but we could still add a spotless check hook till then [diffplug/spotless#178 (comment)](https://github.com/diffplug/spotless/issues/178#issuecomment-351638034)
> > I think @pavolloffay also uses the same plugin in Java Agent https://github.com/hypertrace/javaagent/blob/a339bd5747af6cef4fbc666e641278b61cfbc273/gradle/spotless.gradle
> 
> @skjindal93, are you talking about something like this: https://github.com/thegreystone/action-jmc-spotless ?
> 
> Not sure if this helps. If not we can still write shellscript and use it with dockerfile and create our own action similar to above.
> 
> I believe it will be series of bash commands so we can have basic dockerfile with gradle as in above action and then add bash script as `entrypoint.sh`. would that help?
> 
> We will also need to figure out how GH actions can make a commit if it's going to make changes. Not sure about that.
> 
> Will be happy to work with you if this is significant and can help us in future as well. we can take a stab.

If we just add spotless plugin in the list of applied plugins, spotless exposes a gradle command, which can be used as a separate Github Action as `./gradlew spotlessCheck`. Or, you can make `./gradlew check` depend on `./gradlew spotlessCheck` and since `./gradlew build` already runs `./gradlew check`, it would be automatically taken care of in the CI pipeline

.github/workflows/ci.yml
jcchavezs: I don't think it is about the time it takes, it is mostly about this linter only makes sense if java code has been changed, hence the ignore.

document-store/build.gradle.kts
jcchavezs: We need to fix all files before merge so next person opening a PR in here does not have to deal with the lint mess.

jcchavezs:
I think the gradle plugin to autofix the code is better so we don't add another tool to fix the files. Maybe @pavolloffay can port his work into this repo?

document-store/build.gradle.kts
aaron-steinfeld: Agreed - but aren't these formatting changes going in the wrong direction? They're changing the formatting of files that were previously correctly formatted and after this change are incorrect. As an example, take Datastore.java - https://github.com/hypertrace/document-store/pull/24/files#diff-fcb7b4dcc9273e596438f61590e9eb81e373bec38fa40adec47e2068853fcd3a

The changes switch the block indent from 2 -> 4 (should be 2 https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation ) and separate the lang imports from third party (should be in single block https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing )

aaron-steinfeld:
Agree it would be nice to embed this into the existing `check` lifecycle so there's no ci changes needed and it can also be easily run locally as a standalone gradle task. I'm not sure how spotless works, but one other comment - personally I'd rather CI only fail/succeed based on the result of the check rather than actually change and commit code itself. Certainly nice to have an autofix option available to devs though.

document-store/build.gradle.kts
jcchavezs: Yeah, let me rephrase. This was an attempt to fix all problems but I was fooled by the intellij plugin I think.

jcchavezs:
> Agree it would be nice to embed this into the existing check lifecycle so there's no ci changes needed and it can also be easily run locally as a standalone gradle task. I'm not sure how spotless works, but one other comment - personally I'd rather CI only fail/succeed based on the result of the check rather than actually change and commit code itself. Certainly nice to have an autofix option available to devs though.

Check is good. I think it is too much overhead to only check diffs or diff files. This is linting and checking all files might be just fine. I am also against doing the fix on PR, this should be just a yes/no check and not do anything else. Autofixing on CI is cumbersome experience as 1. actions run in paralell and 2. we need push access 3. users have to download new commits before adding further changes.

pavolloffay:
> I think @pavolloffay also uses the same plugin in Java Agent https://github.com/hypertrace/javaagent/blob/a339bd5747af6cef4fbc666e641278b61cfbc273/gradle/spotless.gradle

Yes the HT javaagent uses the spotless gradle plugin and it works perfectly ❤️ . `./gradlew spotlessApply` formats the code (one can also setup pre-commit hook to run it automagically). The build then checks if the code is formatted correctly - it is done during the unit test execution e.g. `./gradlew check` so no separate CI job is needed.

skjindal93:
@pavolloffay 

> one can also setup pre-commit hook to run it automagically, I am skipping automagic things

This one is still an open issue https://github.com/diffplug/spotless/issues/623, unless we write a fancy bash script to do that

PR #28, opened 2021-01-08
feat: support projections in mongo

skjindal93:
## Description
Adds support for projections in Mongo

### Testing
Added Unit and Integration Tests

### Checklist:

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -15,6 +15,7 @@ import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCursor;
 import com.mongodb.client.model.BulkWriteOptions;
 import com.mongodb.client.model.FindOneAndUpdateOptions;
+import com.mongodb.client.model.Projections;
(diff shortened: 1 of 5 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -218,6 +220,12 @@ private String decodeKey(String key) {
     if (query.getFilter() != null) {
       map = parseQuery(query.getFilter());
     }
+
+    Bson projection = new BasicDBObject();
```
avinashkolluru, line 224: Please add a github issue to track adding support for the same in PostgresCollection

avinashkolluru (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
kotharironak: I am assuming that it won't break any functionality on upgrade if it's not implemented Postgres immediately.

skjindal93: yes. thanks for reminding me that

kotharironak (approved):

PR #30, opened 2021-01-21
Fixing SQLi Vulnerability

kotharironak:
Addresses the issue: #14
- copied the initial work from PR from here - https://github.com/hypertrace/document-store/pull/16 (refer to the discussion on that PR)
- fixes the casting issue where prior all the fields were treated as a string
- adds test case related to different types, and also described fair assumption made on the current fix using a test case (refer to test cases: testWithDifferentDataTypes and testWithDifferentFieldTypes). 
- More discussion on the issue was here - https://github.com/hypertrace/document-store/pull/16#discussion_r542168519


kotharironak:
Verified e2e by changing hypertrace-services with its snapshot jar locally. 
Screenshot 2021-02-04 at 4 03 03 PM
Screenshot 2021-02-04 at 4 05 27 PM

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -482,6 +520,17 @@ private String prepareUpsertDocument(Key key, Document document) throws IOExcept
     return MAPPER.writeValueAsString(jsonNode);
   }
 
+  private String prepareCast(String field, Object value) {
+    String fmt = "CAST (%s AS %s)";
+    if (value instanceof Number) {
+      return String.format(fmt, field, "NUMERIC");
+    } else if (value instanceof Boolean) {
+      return String.format(fmt, field, "BOOLEAN");
+    } else /* default is string */ {
```
avinashkolluru, line 529: What if the value is a List (IN clause case)

avinashkolluru: I couldnt find a test case where the IN clause values are numbers. May be worth checking that case once.

kotharironak: Good point. 

There are two ways to handle this case.
- Option1: For the `IN` operator, it applies equality operator and will not affect sorting order. So, we can cast the values to the string while addObjectParam. It will be the generic solution and can handle mixed types in List
- Option2: As another alternative, we can consider the first element of the List, and consider its type as cast type. It will also be a fair assumption that `IN` or any other operator will have the same type values. So, as consistency, we can have CAST ( <field name> as <type> ) for all non string operator.

I am going ahead with option (2) and will update the PR with change and test case.

kotharironak: Made the changes, and updated the PR. Added this test case for `testInQueryWithNumberField` this specific scenario.

avinashkolluru (approved):

Issue #43, opened 2021-03-17
[Feature Request] support for conditional insert and update


kotharironak:
## Use Case,
The document store's collection API provides an `update` interface that inserts and updates the docuemnt.
There are cases where we would like to insert and update the document if the condition match.
e.g
- Update the document only if the last updated time was lesser than the current time
- Update the document only if the version value is less than the new version value

The above conditional-based update interface will be helpful in client-side optimistic locking for concurrent updates. 

## Proposal
We will have a newer interface that will take Filter as part of the interface which will be added into where clause of an update
e.g
```<Document/boolean> upsert(Key key, Document document, Filter condition) throws IOException;```

## Questions to address (if any)
- explore if the concurrent `insert` will aslo be taken care of.

PR #45, opened 2021-03-19
feat: adds two new apis create and update with for conditional support

kotharironak:
As a part of the requirement - https://github.com/hypertrace/document-store/issues/43 - of supporting the conditional update of document for concurrency requirements, and based on discussion [below](https://github.com/hypertrace/document-store/pull/45/#discussion_r601492847), we are adding two new APIs,
- create API: creates a new document if it doesn't exist
- update API with condition: update an existing document if a condition evaluated to true


document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -16,6 +17,19 @@
...
+   * Upsert (create a new doc or update if one already exists) the given document into the doc store
+   * if condition is evaluated to true. Provides optimistic lock support for concurrency update.
+   *
+   * @param key Unique key of the document in the collection.
+   * @param document Document to be upserted.
+   * @param condition Filter condition to be evaluated, on success update the document
+   * @param isUpsert Optional parameter indicates insert or update, default is true
+   * @return True if success. False otherwise.
+   */
+  boolean upsert(Key key, Document document, Filter condition, @Nullable Boolean isUpsert)
```
surajpuvvada, line 30: wondering if it might be better to create separate APIs instead one for insert and upsert ?

kotharironak: Let's see the use case but here, compared to the existing upsert API, I thought of enhancing it with the optional parameter `isUpsert` if a client wants to explicitly control `insert` and `update`.
```
True indicates if the document doesn't exist, it will insert a new document. And, if a document exists, it will update.
False indicates if the document exists, update it otherwise don't do anything.
```
The default value for isUpsert will be `true`. Let me know, would this suffice?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -121,6 +122,7 @@ public static void shutdown() {
   @MethodSource
   private static Stream<Arguments> databaseContextProvider() {
     return Stream.of(Arguments.of(MONGO_STORE), Arguments.of(POSTGRES_STORE));
+    // return Stream.of(Arguments.of(MONGO_STORE));
```
kotharironak, line 125:
```suggestion
```

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
surajpuvvada: this works as well

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -123,13 +124,36 @@ public boolean upsert(Key key, Document document) throws IOException {
         LOGGER.debug("Write result: " + writeResult.toString());
       }
 
-      return writeResult.getModifiedCount() > 0;
+      return (writeResult.getUpsertedId() != null || writeResult.getModifiedCount() > 0);
```
surajpuvvada, line 127: for insert modifiedcount will be 0 so the upsertId condition is added I assume ?

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
skjindal93, line 32: `@Nullable Boolean` sounds a bit weird. Can we just have `boolean isUpsert`? Or, there is a reason to maintain 3 states? `null`, `true` or `false`

Because, `null` and `true` has the same behavior here

skjindal93: Since we are creating new APIs, I think it would be more cleaner to have separate `create` and `update` APIs. Thoughts?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -82,6 +83,38 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+  public boolean upsert(Key key, Document document, Filter condition, @Nullable Boolean isUpsert)
+      throws IOException {
+    StringBuilder upsertQueryBuilder =
+        new StringBuilder(getUpsertSQL(isUpsert != null ? isUpsert : true));
+
+    String jsonString = prepareUpsertDocument(key, document);
+    Params.Builder paramsBuilder = Params.newBuilder();
+    paramsBuilder.addObjectParam(key.toString());
+    paramsBuilder.addObjectParam(jsonString);
+    paramsBuilder.addObjectParam(jsonString);
```
skjindal93, line 100: accidentally, added the `jsonString` twice? Or, is it deliberate?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -82,6 +83,38 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+
+    if (condition != null && isUpsert == true) {
+      String conditionQuery = parseFilter(condition, paramsBuilder, "d");
```
skjindal93, line 103: Can you please add a comment about the argument `d` here?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -187,20 +220,21 @@ public boolean updateSubDoc(Key key, String subDocPath, Document subDocument) {
   }
 
   @VisibleForTesting
-  protected String parseFilter(Filter filter, Params.Builder paramsBuilder) {
+  protected String parseFilter(Filter filter, Params.Builder paramsBuilder, String aliasPrefix) {
```
skjindal93, line 227: Can you please add the `@Nullable` notation for `aliasPrefix`?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -329,14 +366,15 @@ private StringBuilder prepareFieldAccessorExpr(String fieldName) {
   }
 
   @VisibleForTesting
-  protected String parseCompositeFilter(Filter filter, Params.Builder paramsBuilder) {
+  protected String parseCompositeFilter(
```
skjindal93, line 373: Can you please add the `@Nullable` notation for `aliasPrefix`?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -314,10 +348,13 @@ private String prepareParameterizedStringForList(
     return "(" + collect + ")";
   }
 
-  private StringBuilder prepareFieldAccessorExpr(String fieldName) {
+  private StringBuilder prepareFieldAccessorExpr(String fieldName, String aliasPrefix) {
```
skjindal93, line 355: Can you please add the `@Nullable` notation for `aliasPrefix`?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -1055,6 +1057,188 @@ public void testSearch(String dataStoreName) throws IOException {
...
+
+    // test that document document is inserted if its not exists
+    Iterator<Document> results = collection.search(query);
+    List<Document> documents = new ArrayList<>();
+    while (results.hasNext()) {
+      documents.add(results.next());
+    }
+    Assertions.assertTrue(documents.size() == 0);
+
+    boolean bool =
```
skjindal93, line 1077: nit: `boolean status`

jcchavezs: or `succeed`

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -130,6 +131,30 @@ public boolean upsert(Key key, Document document) throws IOException {
     }
   }
 
+  /** Same as existing upsert method, however, extends the support with condition */
+  @Override
+  public boolean upsert(Key key, Document document, Filter condition, @Nullable Boolean isUpsert)
```
avinashkolluru, line 139: Since we are adding a new api should we also have a better response object encapsulating success/failure, number of docs upserted, any further error desc in case of errors etc etc

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
jcchavezs: Adding this kind of parameters is usually a bad practice as it reflects a bad abstraction for our methods. Let's please have a different method instead of this.

Check 

> Programmer B feels honor-bound to retain the existing abstraction, but since isn't exactly the same for every case, they alter the code to take a parameter, and then add logic to conditionally do the right thing based on the value of that parameter.

in https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction for a illustratory reading :).

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -82,6 +83,42 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+      }
+    }
+
+    try (PreparedStatement preparedStatement =
+        buildPreparedStatement(upsertQueryBuilder.toString(), paramsBuilder.build())) {
+      int result = preparedStatement.executeUpdate();
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Write result: {}", result);
+      }
+      return result > 0;
```
jcchavezs, line 115: what exactly is the result? is it the number of modified records or the id of the update? maybe a more meaningful var name would help?

jcchavezs (changes requested):
I think we can achieve this without the `isUpsert` in a `upsert` method.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
rish691: I feel the methods can be separated on the true/false state of `isUpsert` var, so we would a new `upsert` method accepting filter and other will be an `update` method

jcchavezs: My proposal would be exactly what rishahb described. I don't think we could call an upsert method when we are actually not doing the upsert. Update and upsert are different use cases and my first question would be, do we actually need a `update` function?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/postgres/PostgresDocStoreTest.java
```diff
@@ -98,7 +98,6 @@ public void testInitWithDatabase() {
     }
   }
 
-  @Test
```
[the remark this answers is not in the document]
kotharironak: Aha, looks like due to my manual merge, added back.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: It is deliberate and required.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
kotharironak: No. While adding this, I was still under discussion, should we expose this var at all or not? Removed nullable

kotharironak: > Since we are creating new APIs, I think it would be more cleaner to have separate create and update APIs. Thoughts?
I was in thoughts of not adding additional API w/o having a use case as it will have to maintain. 

Semantically, we already have an upsert method which if a document exists update, and if doesn't exist create one.
We were in a need of having one which provides optimistic locking for concurrency. So, initially, I was thinking of just extending it with the condition.
```
boolean upsert(Key key, Document document, Filter condition)
```
@Suraj, will this suffice? If not, I will create two separate APIs.

Create API: Create a document it doesn't exist
```
boolean create(Key key, Document document)
```
Update API: update the document if the condition met
```
boolean update(Key key, Document document, Filter condition)
```

kotharironak: Discussed offline with @jcchavezs.
The API 
```
boolean upsert(Key key, Document document, Filter condition)
```
solves the use cases required for concurrency, however, it bit inconsistent. For the insert, it ignores the condition. But, the client can expect otherwise that if the condition is not met, it should not create the document. So, we will go ahead with the second proposal of having two separate APIs - create and update with the condition.

```
boolean create(Key key, Document document)
boolean update(Key key, Document document, Filter condition)
```

surajpuvvada: Yes this is much easier to follow

aaron-steinfeld:
So I actually just ran into (almost) this exact same requirement to support concurrent entity updates. The one aspect in the entity case that I think is missing is the actual current state. This is kind of a subset of avinash's comment: https://github.com/hypertrace/document-store/pull/45#discussion_r600404881

Basically, if I attempt to update something and it's rejected because the update is now stale, it would be helpful to have the current state as part of the response. Of course, we can fetch it after if that's not feasible, but it would save a round trip to return it directly (in other words, this is the same requirement that brought in the `upsertAndReturn` API).

kotharironak:
> So I actually just ran into (almost) this exact same requirement to support concurrent entity updates. The one aspect in the entity case that I think is missing is the actual current state. This is kind of a subset of avinash's comment: [#45 (comment)](https://github.com/hypertrace/document-store/pull/45#discussion_r600404881)
> 
> Basically, if I attempt to update something and it's rejected because the update is now stale, it would be helpful to have the current state as part of the response. Of course, we can fetch it after if that's not feasible, but it would save a round trip to return it directly (in other words, this is the same requirement that brought in the `upsertAndReturn` API).

As of now, it looked to me that both MongoDB and Postgres, have support for returning document if an operation is successful, but not on failure. Postgres support it using the `RETURNING` expression which is an extension to PostgresSQL - https://www.postgresql.org/docs/9.5/sql-insert.html

aaron-steinfeld:
> > So I actually just ran into (almost) this exact same requirement to support concurrent entity updates. The one aspect in the entity case that I think is missing is the actual current state. This is kind of a subset of avinash's comment: [#45 (comment)](https://github.com/hypertrace/document-store/pull/45#discussion_r600404881)
> > Basically, if I attempt to update something and it's rejected because the update is now stale, it would be helpful to have the current state as part of the response. Of course, we can fetch it after if that's not feasible, but it would save a round trip to return it directly (in other words, this is the same requirement that brought in the `upsertAndReturn` API).
> 
> As of now, it looked to me that both MongoDB and Postgres, have support for returning document if an operation is successful, but not on failure. Postgres support it using the `RETURNING` expression which is an extension to PostgresSQL - https://www.postgresql.org/docs/9.5/sql-insert.html

Hmm - ok we can disregard. My use case actually doesn't require this as I built it out because entity upserts are merged - I have to fetch the value first to do that merge, so I'm just enforcing the condition at that time before proceeding to either create or update.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
kotharironak: Yes. But, this is not needed now.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
kotharironak: Done. Added two new APIs.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -93,5 +93,25 @@
...
+   * create a new document if one doesn't exists with the key
+   *
+   * @param key Unique key of the document in the collection.
+   * @param document Document to be created.
+   * @return an instance of {@link DocStoreResult}
+   */
+  DocStoreResult create(Key key, Document document) throws IOException;
+
+  /**
+   * Update an existing document if condition is evaluated to true. Conditional will help in
```
avinashkolluru, line 106:
```suggestion
   * Update an existing document if condition is evaluated to true. Condition will help in
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -82,6 +84,68 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+
+    try (PreparedStatement preparedStatement =
+        buildPreparedStatement(upsertQueryBuilder.toString(), paramsBuilder.build())) {
+      int result = preparedStatement.executeUpdate();
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Write result: {}", result);
+      }
+      DocStoreResult.Builder resultBuilder = DocStoreResult.Builder.newBuilder();
+      resultBuilder.setSuccess(result > 0).setResultCount(result);
+      if (result <= 0) {
```
avinashkolluru, line 115: I'm not sure if result=0 can be considered a failure. Isn't it something for the consumers/clients to decide based on the business logic/use case

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -123,13 +125,66 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+      DocStoreResult.Builder resultBuilder = DocStoreResult.Builder.newBuilder();
+      BasicDBObject basicDBObject = this.prepareInsert(key, document);
+      InsertOneResult insertOneResult = collection.insertOne(basicDBObject);
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Create result: " + insertOneResult.toString());
+      }
+      boolean success = insertOneResult.getInsertedId() != null;
+      resultBuilder.setSuccess(success).setResultCount(1);
+      if (!success) {
+        resultBuilder.setFailureMessage("Failed to create document with key:" + key.toString());
```
surajpuvvada, line 179: should the original error message be captured and included here ? for example - duplicate error or something else ? The doc for insertOne actually throws an exception - which I'm assuming InsertOneResult is capturing ?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: No longer needed.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: On condition fail, and the document is not updated, it is not an error. So, the API will not throw any error. And as per the java doc of prepared stmt, it returns `0` if there is no modification.

> @return either (1) the row count for SQL Data Manipulation Language (DML) statements or (2) 0 for SQL statements that return nothing

So, if we keep the `success` status like all othersAPIs, we have to determine using `result`. If we don't we can only keep `resultCount` in DocStoreResult.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
kotharironak: If there is an exception, that will be wrapped into IOException and thrown to the client. The duplicate error is an exception scenario. Here, we are handling the case where there was no exception.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: As discussed with @avinashkolluru, we will return the following response object and client will make a call accordingly.
For create API:
```
public class CreateResult {
  private long createdCount;
}
```
For Update API
```
public class UpdateResult {
  private long updatedCount;
}
```

kotharironak: @avinashkolluru I have addressed this.

surajpuvvada (approved):
lgtm

avinashkolluru (approved):
Approved but i think there are better ways to unit test multi threaded code. The test you have currently is not guaranteed to execute in a predictable manner because we cant control the interleaving of threads spawned. 

https://www.baeldung.com/java-testing-multithreaded some better techniques of doing this.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -1055,6 +1061,188 @@ public void testSearch(String dataStoreName) throws IOException {
...
+      documents.add(results.next());
+    }
+    Assertions.assertTrue(documents.size() == 1);
+    Map<String, Object> doc = OBJECT_MAPPER.readValue(documents.get(0).toJson(), Map.class);
+    Assertions.assertEquals(resultMap.get(SUCCESS).get(0).getTestValue(), (int) doc.get("size"));
+  }
+
+  @ParameterizedTest
+  @MethodSource("databaseContextProvider")
+  public void testUpdateWithMultipleThreads(String dataStoreName) throws Exception {
```
jcchavezs, line 1092: Should we include a test for concurrently update where the first one gets updated the rest fail? Maybe it is already in but I can't not easy identify such test from reading them.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/CreateUpdateTestThread.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.hypertrace.core.documentstore.Collection;
+import org.hypertrace.core.documentstore.CreateResult;
+import org.hypertrace.core.documentstore.Document;
+import org.hypertrace.core.documentstore.Filter;
+import org.hypertrace.core.documentstore.Filter.Op;
+import org.hypertrace.core.documentstore.SingleValueKey;
+import org.hypertrace.core.documentstore.UpdateResult;
+
+public class CreateUpdateTestThread extends Thread {
```
jcchavezs, line 13: should this be final? Also, could we elaborate why this class is needed?

document-store/src/main/java/org/hypertrace/core/documentstore/UpdateResult.java
```diff
@@ -0,0 +1,16 @@
...
+ * */
+public class UpdateResult {
+  private long updatedCount;
+
+  public UpdateResult(long updatedCount) {
+    this.updatedCount = updatedCount;
+  }
+
+  public long getUpdatedCount() {
+    return updatedCount;
```
jcchavezs, line 14: Is this absolutely necessary? Also, can't we make it lazy? In general I think this result should be an interface as some databases retrieve this values for free and others have to perform operations so I'd leave this up to the implementation.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -643,6 +698,18 @@ private String prepareCast(String field, Object value) {
...
 
+  private String getInsertSQL() {
+    return String.format(
+        "INSERT INTO %s (%s,%s) VALUES( ?, ? :: jsonb)",
+        collectionName, ID, DOCUMENT, ID, DOCUMENT);
+  }
+
+  private String getUpdateSQL() {
+    return String.format(
+        "UPDATE %s AS d SET (%s, %s) = ( ?, ? :: jsonb) ",
```
jcchavezs, line 709: just curious, is `AS d` necessary?

kotharironak: No. Oops, I thought, I had removed it.

document-store/src/main/java/org/hypertrace/core/documentstore/CreateResult.java
```diff
@@ -0,0 +1,16 @@
+package org.hypertrace.core.documentstore;
+
+/*
+ * Represent the result object for CREATE operation of document store APIs.
+ * */
+public class CreateResult {
```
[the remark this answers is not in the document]
kotharironak: we are already throwing exceptions on error. This will be useful if we want to use it along with bulk create API.

document-store/src/main/java/org/hypertrace/core/documentstore/UpdateResult.java
kotharironak: > Also, can't we make it lazy?
Didn't get it.

document-store/src/main/java/org/hypertrace/core/documentstore/CreateResult.java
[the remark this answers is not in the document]
jcchavezs: having to check 0vs1 in the CreateResult only adds one extra step to know if something was successful (`if (r.getCreatedCount() == 0)` vs `if(r.succeeded())`, could we have something like

```java
class CreateResult {
    public boolean succeed()
}
class BulkCreateResult extends CreateResult {
    public int getCreatedCount()
}
```

In update would be the same, the fact that no document matched the criteria doesn't mean the update operation failed. Some users would care about the number of affected, some others won't. It is the same criteria for the DB apis where some only execute the update and others look at the `affected rows` value. If they don't couple them together, should we do it?

cc @avinashkolluru

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: Yes. this one is doing the same.
- create one document
- `5` threads trying to update with the condition, and only one success, and for same I am using test class `CreateUpdateTestThread`

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/CreateUpdateTestThread.java
kotharironak: will make it final. Used in concurrency tests - e.g `testUpdateWithMultipleThreads`

document-store/src/main/java/org/hypertrace/core/documentstore/CreateResult.java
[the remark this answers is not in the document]
kotharironak: What would be `BulkCreateResult`'s succeed method returns on partial create? Say, if we requested to create N documents and the only subset of it get created. So, either we go with response object/per API or with a smaller set as per each CURD operation. As I was suggesting the second option - `per CURD` operation, we went ahead with the above class design.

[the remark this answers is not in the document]
kotharironak: Addressed this, we will go with a response object per API.

avinashkolluru (dismissed):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -123,13 +125,55 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+  /**
+   * Update an existing document if condition is evaluated to true. Conditional will help in
+   * providing optimistic locking support for concurrency update.
+   */
+  @Override
+  public org.hypertrace.core.documentstore.UpdateResult update(
+      Key key, Document document, Filter condition) throws IOException {
+    try {
+      Map<String, Object> conditionMap =
+          condition != null ? parseQuery(condition) : new HashMap<>();
```
jcchavezs, line 144:
```suggestion
          condition == null ? new HashMap<>() : parseQuery(condition);
```

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -123,13 +125,55 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+      UpdateOptions options = new UpdateOptions().upsert(false);
+
+      UpdateResult writeResult =
+          collection.updateOne(conditionObject, this.prepareUpsert(key, document), options);
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Update result: " + writeResult.toString());
+      }
+      return new org.hypertrace.core.documentstore.UpdateResult(writeResult.getModifiedCount());
+    } catch (Exception e) {
+      LOGGER.error("Exception updating document. key: {} content:{}", key, document, e);
```
jcchavezs, line 156:
```suggestion
      LOGGER.error("Exception updating document. key: {} content: {}", key, document, e);
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -643,6 +698,17 @@ private String prepareCast(String field, Object value) {
     }
   }
 
+  private String getInsertSQL() {
+    return String.format(
+        "INSERT INTO %s (%s,%s) VALUES( ?, ? :: jsonb)",
+        collectionName, ID, DOCUMENT, ID, DOCUMENT);
+  }
+
+  private String getUpdateSQL() {
```
jcchavezs, line 707: I think this can perfectly be static but I also wonder if we need to run the format every time, can't we format the SQL statement on construct and reuse it?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -82,6 +85,58 @@ public boolean upsert(Key key, Document document) throws IOException {
...
+
+    try (PreparedStatement preparedStatement =
+        buildPreparedStatement(upsertQueryBuilder.toString(), paramsBuilder.build())) {
+      int result = preparedStatement.executeUpdate();
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Write result: {}", result);
+      }
+      return new UpdateResult(result);
+    } catch (SQLException e) {
+      LOGGER.error("SQLException inserting document. key: {} content:{}", key, document, e);
```
jcchavezs, line 116:
```suggestion
      LOGGER.error("SQLException inserting document. key: {} content: {}", key, document, e);
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: It uses `collectionName` (table) and its per collection object - `PostgresCollection`. So, we can optimize a bit. I will raise a new refactor PR for all the similar SQL method statements.

jcchavezs (approved):

avinashkolluru (approved):

PR #51, opened 2021-04-27
Fixes #50: Handle sorting of documents on multiple attributes

### LINKED ISSUE #50: Sorting documents by multiple attributes is not working correctly
## Problem
When we provide multiple `OrderBy`s in a query, it sometimes returns the documents in an incorrect order. For example, if we try to sort the documents based on `field1` and then `field2`, it works incorrectly in some cases:
`query.addAllOrderBys(List.of(new OrderBy("field1", true), new OrderBy("field2", false)));`

This is tested using MongoCollection.

## Cause
The orderBy list is converted to a HashMap which may not preserve the order of the keys(as in the list) -
https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java#L365
So, in the above example, it can happen that the sorting happens first on `field2` and then on `field1`.

GurtejSohi:
## Description
Fixes https://github.com/hypertrace/document-store/issues/50 by using `LinkedHashMap` instead of `HashMap` for `orderByMap` to preserve the order of keys (as in the list).

### Testing
Added unit test

### Checklist:

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
@@ -29,6 +29,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -361,9 +362,7 @@ public class MongoCollection implements Collection {
     }
 
     if (!query.getOrderBys().isEmpty()) {
-      Map<String, Object> orderbyMap = new HashMap<>();
-      parseOrderByQuery(query.getOrderBys(), orderbyMap);
-      BasicDBObject orderBy = new BasicDBObject(orderbyMap);
+      BasicDBObject orderBy = new BasicDBObject(parseOrderByQuery(query.getOrderBys()));
       cursor.sort(orderBy);
     }
 
@@ -409,10 +408,13 @@ public class MongoCollection implements Collection {
     return true;
   }
 
-  private void parseOrderByQuery(List<OrderBy> orderBys, Map<String, Object> orderbyMap) {
+  @VisibleForTesting
+  LinkedHashMap<String, Object> parseOrderByQuery(List<OrderBy> orderBys) {
+    LinkedHashMap<String, Object> orderByMap = new LinkedHashMap<>();
     for (OrderBy orderBy : orderBys) {
-      orderbyMap.put(orderBy.getField(), (orderBy.isAsc() ? 1 : -1));
+      orderByMap.put(orderBy.getField(), (orderBy.isAsc() ? 1 : -1));
     }
+    return orderByMap;
   }
 
   @VisibleForTesting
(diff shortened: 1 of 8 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -409,10 +408,13 @@ public boolean deleteAll() {
     return true;
   }
 
-  private void parseOrderByQuery(List<OrderBy> orderBys, Map<String, Object> orderbyMap) {
+  @VisibleForTesting
+  LinkedHashMap<String, Object> parseOrderByQuery(List<OrderBy> orderBys) {
```
aaron-steinfeld, line 412: nit: perhaps a query parser util/class would better follow SRP and remove the need for `@VisibleForTesting`, everyone's favorite antipattern 😉

aaron-steinfeld (commented):
lgtm

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -409,10 +408,13 @@ public boolean deleteAll() {
...
+    LinkedHashMap<String, Object> orderByMap = new LinkedHashMap<>();
     for (OrderBy orderBy : orderBys) {
-      orderbyMap.put(orderBy.getField(), (orderBy.isAsc() ? 1 : -1));
+      orderByMap.put(orderBy.getField(), (orderBy.isAsc() ? 1 : -1));
```
kotharironak, line 415: Can you extend the test `testOffsetAndLimitOrderBy` with multiple orderBy in [DocStoreTest.java](https://github.com/hypertrace/document-store/blob/081cf1b5f9433e22d1a8736d01e1cc85dd663109/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java#L827), so, it can verify Postgres side too. However, I have checked the Postgres side impl, it looks fine for multiple orderby.

kotharironak (commented):
overall, lgtm.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
GurtejSohi: Moved the query parsing methods into QueryParser classes - one for mongo and another for postgres.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
GurtejSohi: Sure, extended that test to add multiple orderBys.
Also, added a unit test for `parseOrderBys` method for Postgres.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryParser.java
```diff
@@ -0,0 +1,103 @@
...
+          throw new UnsupportedOperationException(
+              String.format("Boolean operation:%s not supported", op));
+      }
+    } else {
+      Filter.Op op = filter.getOp();
+      Object value = filter.getValue();
+      Map<String, Object> map = new HashMap<>();
+      switch (op) {
+        case EQ:
+          map.put(filter.getFieldName(), value);
```
aaron-steinfeld, line 46: Looks like this just got moved. The create/put/break/return kills me but that's probably scope creep.

aaron-steinfeld (approved):

kotharironak (approved):

PR #52, opened 2021-06-02
fix: postgres iterator interface to handle any order

kotharironak:
## Description
Fix for the issue - https://github.com/hypertrace/hypertrace/issues/248

As recently, we have added support for time-agnostic API support and the corresponding GraphQL query was failing.

```
{
  entities(
    type: SERVICE
    limit: 1
    between: {startTime: "2021-06-01T14:59:41.614Z", endTime: "2021-06-01T15:59:41.614Z"}
    filterBy: [{operator: EQUALS, value: "b89ddf9b-c582-3d93-9006-a4b3024cf5a5", type: ID, idType: SERVICE}]
    includeInactive: true
  ) {
    results {
      id
      name: attribute(key: "name")
      __typename
    }
    __typename
  }
}
```

As the flag `includeInactive: true` along with attribute data source, triggers different workflow for fetching from the document store ( mongo/postgres) or OLAP store. In the previous release, it was fetched from pinot. But, in the latest release, it is fetched from the document store. 

There is a bug in Postgres hasNext interface implementation which assumes a fixed order of fetch as pattern `[hasNext, next, hasNext, next]`. However, a client can call in any order like `[hasNext, hasNext, next]` that is the exact thing happening in the new flow (EntityQueryServiceImpl at [first](https://github.com/hypertrace/entity-service/blob/f8151b08d3a2e35f9ad3e91b9a26774378cc0cae/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java#L113) hasNext and [second](https://github.com/hypertrace/entity-service/blob/f8151b08d3a2e35f9ad3e91b9a26774378cc0cae/entity-service-impl/src/main/java/org/hypertrace/entity/query/service/EntityQueryServiceImpl.java#L125) hasNext). 

The below PR fix the issue. As `isLast` and other ResultSet API are costly, we are fixing them at our end

### Testing
- Added integration test.
- verified locally with docker-compose setup

output:
Screenshot 2021-06-01 at 6 02 23 PM

### Checklist:
- [ ] My changes generate no new warnings

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
@@ -493,7 +493,7 @@ public class PostgresCollection implements Collection {
 
     private final ObjectMapper MAPPER = new ObjectMapper();
     private ResultSet resultSet;
-    private boolean didNext = false;
+    private boolean cursorMovedForward = false;
     private boolean hasNext = false;
 
     public PostgresResultIterator(ResultSet resultSet) {
@@ -503,9 +503,9 @@ public class PostgresCollection implements Collection {
     @Override
     public boolean hasNext() {
       try {
-        if (!didNext) {
+        if (!cursorMovedForward) {
           hasNext = resultSet.next();
-          didNext = true;
+          cursorMovedForward = true;
         }
         return hasNext;
       } catch (SQLException e) {
@@ -517,10 +517,11 @@ public class PostgresCollection implements Collection {
     @Override
     public Document next() {
       try {
-        if (!didNext) {
+        if (!cursorMovedForward) {
           resultSet.next();
         }
-        didNext = false;
+        // reset the cursorMovedForward state, if it was forwarded in hasNext.
+        cursorMovedForward = false;
         return prepareDocument();
       } catch (IOException | SQLException e) {
         return JSONDocument.errorDocument(e.getMessage());
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
@@ -1,5 +1,6 @@
 package org.hypertrace.core.documentstore.postgres;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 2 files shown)


JBAhire (dismissed):

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -500,7 +503,11 @@ public PostgresResultIterator(ResultSet resultSet) {
     @Override
     public boolean hasNext() {
       try {
-        return resultSet.next();
+        if (!didNext) {
+          hasNext = resultSet.next();
+          didNext = true;
```
rish691, line 508: why is the `didNext` state changing on the hasNext call?
I assume that the `didNext` var implies if `next` has been called on the current iterator pointer or not?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -491,7 +492,9 @@ public void drop() {
   static class PostgresResultIterator implements Iterator {
 
     private final ObjectMapper MAPPER = new ObjectMapper();
-    ResultSet resultSet;
+    private ResultSet resultSet;
+    private boolean didNext = false;
```
rish691, line 496: can we rename this var?

kotharironak: I have renamed it to `cursorMovedForward`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: There is no support for `hastNext` in the underlying `ResultSet` interface. They do have support for functions like `isLast` for traversal and position checking, but some of them are costly.  Secondly, it has a directional concept on the cursor - e.g `TYPE_FORWARD_ONLY`. So, I consider implementing `hashNext` and `next` using `ResultSet's next` only.

As, hashNext has to say that there are more items, but to answer, it advances a cursor, I have to maintain the state.

rish691 (approved):

avinashkolluru (approved):

PR #55, opened 2021-10-20
Added support for bulk operations on array value

saxenakshitiz:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -333,6 +334,76 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        break;
+      default:
+        throw new UnsupportedOperationException("Unknown operation : " + request.getOperation());
+    }
+
+    List<UpdateManyModel<BasicDBObject>> bulkWriteUpdate =
+        List.of(
+            new UpdateManyModel(
+                selectionCriteriaForKeys(request.getKeys()), operationObject, new UpdateOptions()));
+    BulkWriteResult writeResult = collection.bulkWrite(bulkWriteUpdate);
```
aaron-steinfeld, line 377: So we had talked separately about introducing a new upsert api that supports transactionality. Would it make sense to start that here? My concern is that every new upsert use case just gets its own little individual api, then we start getting combinations and the api keeps growing in a crazy way. I'm basically imagining one flexible api that encompasses all of our upsert needs - accepting one or more operations, with described transactionality, merging, sub docs, etc. That API doesn't need to be built out in a day, but if we could start building new use cases like this one into it, that would be awesome.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -225,6 +226,11 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
     }
   }
 
+  @Override
+  public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request) {
+    throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
aaron-steinfeld: ok, separate PR is fine.

document-store/src/main/java/org/hypertrace/core/documentstore/BulkArrayValueUpdateRequest.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public Collection<Document> getSubDocuments() {
+    return subDocuments;
+  }
+
+  public enum Operation {
+    ADD,
+    REMOVE,
+    SET
```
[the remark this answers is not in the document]
aaron-steinfeld: I think I value having less independent ways to do something vs optimizing the request brevity for a particular case, but I guess it's complexity on the caller to have to call different APIs depending on the operation (another reason to consolidate to one api to rule them all!)

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -265,8 +265,15 @@ private BasicDBObject prepareDocument(Key key, Document document, long now)
   /** Updates auto-field lastUpdatedTime when sub doc is updated */
   @Override
   public boolean updateSubDoc(Key key, String subDocPath, Document subDocument) {
+    String jsonString = subDocument.toJson();
     try {
-      BasicDBObject dbObject = getSanitizedObject(subDocument);
+      JsonNode jsonNode = MAPPER.readTree(jsonString);
```
aaron-steinfeld: What failed when using the method version? Do we have different sanitization in different places?

aaron-steinfeld: This seems concerning to me - what's different? which one is correct?

saxenakshitiz: Missed the subdoc path. Fixed it.

avinashkolluru (approved):

PR #57, opened 2021-11-05
Add aggregation support with an implementation for MongoDB

suresh-prakash:
## Description
Draft changes for design review.
Will add unit tests once the design is accepted.

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### Documentation


document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import org.apache.commons.lang3.StringUtils;
+import org.hypertrace.core.documentstore.expression.Expression;
+
+@Data
+@AllArgsConstructor
+public class GroupBy {
+  // TODO: Implement a builder for easily constructing the "group by" clause
+  private Expression key; // The key field whose distinct values would be returned
```
avinashkolluru, line 13: The use of Expression seems to be very specific to GroupBy. 
Based on how its integrated, we can't do something like a select round(key) from collection.

So, should we rather enhance the Query class to allow for expressions in selections, orderby, filters too? This effectively means we cant reuse the current Query class in a backward compatible way. So may be a new method/api altogether.
cc: @aaron-steinfeld

document-store/src/main/java/org/hypertrace/core/documentstore/expression/UnaryOperatorExpression.java
```diff
@@ -0,0 +1,18 @@
+package org.hypertrace.core.documentstore.expression;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+@EqualsAndHashCode
+public class UnaryOperatorExpression implements OperatorExpression {
```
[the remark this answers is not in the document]
suresh-prakash: Right! We could easily generalise it. Will change the implementation.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.List;
+import lombok.Data;
+import org.apache.commons.lang3.StringUtils;
+import org.hypertrace.core.documentstore.expression.Expression;
+
+@Data
+public class GroupBy {
+  private Expression key; // The key field whose distinct values would be returned
+  private List<GroupingSpec> groupingSpecs; // The list of aggregations required
```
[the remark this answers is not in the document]
suresh-prakash: Yes, you're right in the sense that these are more like (special) projections.
For Mongo, these need to specified along with the "group" clause. Ref.: https://docs.mongodb.com/upcoming/reference/operator/aggregation/group/.

However, since this interface is not-specific to any DB/implementation, we could consider that. I will try to move it under selections. Doing so might need additional code changes when we increment this library version since selections today are simple strings.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
suresh-prakash: Yeah. For the first iteration, I only implemented expressions under GroupBy. I totally agree with you that we need to make it more generic (though it might break the existing compatibility).

document-store/build.gradle.kts
```diff
@@ -8,13 +8,18 @@ plugins {
 
 dependencies {
   api("com.typesafe:config:1.3.2")
+  annotationProcessor("org.projectlombok:lombok:1.18.22")
+  implementation("org.projectlombok:lombok:1.18.22")
```
aaron-steinfeld, line 12: lombok is compileOnly, we never need it at runtime (here and in test)

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
```diff
@@ -0,0 +1,35 @@
+package org.hypertrace.core.documentstore;
+
+import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import org.apache.commons.lang3.StringUtils;
+import org.hypertrace.core.documentstore.expression.Expression;
+
+@Data
```
aaron-steinfeld, line 9: Can this be a value (and the constructor is implied)?

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
```diff
@@ -0,0 +1,35 @@
...
+@AllArgsConstructor
+public class GroupBy {
+  // TODO: Implement a builder for easily constructing the "group by" clause
```
aaron-steinfeld, line 12: Can add `@Builder`

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
aaron-steinfeld: This class seems to encompass both selections and group bys, which is difficult to follow. I'd suggest starting with an assumed new API, let's see where it goes and then we can figure out if there's any reasonable way to make it compatible.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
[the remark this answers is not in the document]
aaron-steinfeld: +1, should have read this before posting my earlier comment :)

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
aaron-steinfeld: Also, we should support (or leave room in the API to support), multiple levels of group by.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
```diff
@@ -0,0 +1,35 @@
...
+    LAST,
+    MIN,
+    MAX,
+    COUNT,
+    SUM,
+    AVERAGE,
+  }
+
+  @Override
+  public String toString() {
```
aaron-steinfeld, line 27: toString is provided by lombok

document-store/src/main/java/org/hypertrace/core/documentstore/GroupingSpec.java
```diff
@@ -0,0 +1,20 @@
+package org.hypertrace.core.documentstore;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import org.hypertrace.core.documentstore.expression.Expression;
+
+@AllArgsConstructor
+@Getter
+@EqualsAndHashCode
```
aaron-steinfeld, line 10: Value

document-store/src/main/java/org/hypertrace/core/documentstore/GroupingSpec.java
```diff
@@ -0,0 +1,20 @@
...
+public class GroupingSpec {
+  private Expression expression; // The expression to be grouped
+  private GroupBy.Accumulator accumulator; // The accumulator function
+  private String alias; // Name of the accumulation result
+
+  @Override
+  public String toString() {
+    return String.format("%s(%s) AS %s", accumulator.name(), expression.toString(), alias);
```
aaron-steinfeld, line 18: remove and use lombok-provided

document-store/src/main/java/org/hypertrace/core/documentstore/expression/BinaryOperatorExpression.java
```diff
@@ -0,0 +1,19 @@
+package org.hypertrace.core.documentstore.expression;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@AllArgsConstructor
```
aaron-steinfeld, line 7: same lombok comments apply throughout.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/BinaryOperator.java
```diff
@@ -0,0 +1,11 @@
+package org.hypertrace.core.documentstore.expression;
+
+public enum BinaryOperator {
+  ADD,
+  SUBTRACT,
+  MULTIPLY,
+  DIVIDE,
+  REMINDER,
```
aaron-steinfeld, line 8: Guessing this is supposed to be remainder. It's easier to add things than remove them, so I would start with the bare minimum here. Maybe just the first four?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/UnaryOperator.java
```diff
@@ -0,0 +1,11 @@
+package org.hypertrace.core.documentstore.expression;
+
+public enum UnaryOperator {
```
aaron-steinfeld, line 3: Here too, minimize what we expose unless necessary.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/LiteralExpression.java
```diff
@@ -0,0 +1,17 @@
+package org.hypertrace.core.documentstore.expression;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+@EqualsAndHashCode
+public class LiteralExpression implements Expression {
```
aaron-steinfeld, line 10: How is this different than a constant expression?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/UnaryOperatorExpression.java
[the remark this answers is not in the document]
aaron-steinfeld: Some of this support was already in the existing implementation right? Are we re-implementing it because it wasn't strongly typed, isn't usable here or some other shortcoming?

aaron-steinfeld (commented):
Only went through the API - especially for a big set of functionality like this, please always get the api reviewed and agreed on before working on any implementation, it's just burning cycles otherwise.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/UnaryOperatorExpression.java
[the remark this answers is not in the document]
suresh-prakash: This is for implementing GroupBy.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
suresh-prakash: What exactly do you mean by "multiple levels" of group by?
Did you mean, "GROUP BY colA, colB, colC"?

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
suresh-prakash: I didn't know this annotation before. Thanks for the tip. :)

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
suresh-prakash: +1 for this tip too. :)

document-store/src/main/java/org/hypertrace/core/documentstore/expression/BinaryOperator.java
suresh-prakash: Nice catch. Sure.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/LiteralExpression.java
suresh-prakash: Literal expressions are for variables like 
`SELECT MIN(col1) FROM table`
whereas constant expressions are for specifying constants like
`SELECT SUM(1) FROM table`

Specifically, Mongo requires literals to be prefixed with "$", whereas constants should be used without any prefix.

suresh-prakash:
> Only went through the API - especially for a big set of functionality like this, please always get the api reviewed and agreed on before working on any implementation, it's just burning cycles otherwise.

Makes sense. Will make sure to get the API reviewed first, before adding business logic.
I have just created a separate PR with the new API changes: https://github.com/hypertrace/document-store/pull/58/files

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -275,6 +275,11 @@ public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest re
     return Collections.emptyIterator();
   }
 
+  @Override
+  public Iterator<Document> aggregate(Query query) {
```
kotharironak, line 279: will this break postgres HT setup if this support is not implanted on Postgres side?

suresh-prakash: It would not as we are introducing a new API, which we would be consuming later.

document-store/src/main/java/org/hypertrace/core/documentstore/GroupBy.java
aaron-steinfeld: yep, exactly

suresh-prakash: Got it! Please refer to https://github.com/hypertrace/document-store/pull/58 for the API changes. Multiple grouping levels is supported in these API changes. However, need to figure out how to implement this for MongoDB.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/LiteralExpression.java
suresh-prakash: Makes sense. I will rename LiteralExpression to IdentifierExpression in the new API PR (https://github.com/hypertrace/document-store/pull/58)

suresh-prakash:
Part of the code moved to the PR https://github.com/hypertrace/document-store/pull/59

PR #58, opened 2021-11-08
Introduce GenericQuery API as a potential replacement for Query class

suresh-prakash:
## Description
1. Divide the expression types into 6 categories namely,
   * Constant
   * Literal
   * Arithmetic
   * Aggregate
   * Relational
   * Logical
2. Tag each expression type based on the supported clause(s) using the following marker interfaces
   Filterable: Relational, Logical
   Projectable: Constant, Literal, Arithmetic, Aggregate
   Groupable: Literal, Arithmetic
   Sortable: Literal, Arithmetic, Aggregate
3. Integrate these interfaces with necessary wrappers into the GenericQuery class, along with other fields
4. Each expression and the whole query can be constructed using builders or static constructors based on the
   number of fields

### Testing
Unit tests are added. Integration tests will be added in a separate PR.

### Checklist:

### Documentation


document-store/src/main/java/org/hypertrace/core/documentstore/expression/AggregateExpression.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Expression representing aggregation in a query.
+ *
+ * <p>Example: SUM(col1) can be constructed as <code>
+ *    AggregatorExpression.of(Aggregator.SUM, LiteralExpression.of("col1"));
+ * </code>
+ */
+@Value(staticConstructor = "of")
+public class AggregateExpression implements Projectable, Sortable {
```
aaron-steinfeld, line 13: Marker interfaces are often considered an antipattern (here's the first article I found talking about why: https://blog.ndepend.com/marker-interface-isnt-pattern-good-idea/ )- it seems like what we're really going for here is trying to get around the fact that java doesn't have a strong concept of unions. Either putting methods on the interfaces or using a different pattern (such as how oneof is represented in the java generated by protoc) would likely be a better fit.

aaron-steinfeld: Is it possible for something to be projectable but not sortable? Or vice versa? If not, it may make sense to merge the two.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Aggregator.java
```diff
@@ -0,0 +1,7 @@
+package org.hypertrace.core.documentstore.expression;
+
+public enum Aggregator {
+  COUNT,
```
aaron-steinfeld, line 4: distinct count is often more interesting than count and many of our use cases are built on it - probably want that too.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/ArithmeticExpression.java
```diff
@@ -0,0 +1,31 @@
...
+ *   ArithmeticExpression.builder()
+ *     .operator(ArithmeticOperator.SUBTRACT)
+ *     .operand(LiteralExpression.of("A"))
+ *     .operand(ConstantExpression.of(5))
+ *     .build();
+ * </code>
+ */
+@Value
+@Builder
+public class ArithmeticExpression implements Groupable, Projectable, Sortable {
```
aaron-steinfeld, line 28: I'm not sure arithmetic is the right name here. This is really representing any non-aggregate expression as I read it (e.g. `substring(my_col, 0, 3)`)

document-store/src/main/java/org/hypertrace/core/documentstore/expression/ArithmeticOperator.java
```diff
@@ -0,0 +1,15 @@
+package org.hypertrace.core.documentstore.expression;
+
+public enum ArithmeticOperator {
+  // Unary operations
+  ABS,
+  FLOOR,
+
+  // Binary operations
+  SUBTRACT, // Minuend and then subtrahend
```
aaron-steinfeld, line 9: both of these could be considered n-ary in the same way as add and multiply. e.g. `A - B - C` is just as valid as `A + B + C`

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Filterable.java
```diff
@@ -0,0 +1,7 @@
+package org.hypertrace.core.documentstore.expression;
+
+/**
+ * A marker interface to represent that the expression is filterable and can be used in either the
+ * WHERE clause or the HAVING clause of the query.
```
aaron-steinfeld, line 5: where and having would be different, I believe. Having would need an aggregate expression, while where would need to be non-aggregated (I really need a better term for that).

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Filterable.java
```diff
@@ -0,0 +1,7 @@
...
+ */
+public interface Filterable {}
```
aaron-steinfeld, line 7: Independent of the other comments about marker interfaces and aggregates, I'm not sure these names are quite right. `Filterable`, in my mind, implies something can be filtered. Here, we're using this to actually say it does the filtering (maybe `Predicate` would be a better fit for this one? Same idea applies to others - more generically, could do things like `FilteringExpression`)

document-store/src/main/java/org/hypertrace/core/documentstore/expression/LiteralExpression.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Value;
+
+/**
+ * Expression representing either a literal (or a column name)
+ *
+ * <p>Example: LiteralExpression.of("col1");
+ */
+@Value(staticConstructor = "of")
+public class LiteralExpression implements Groupable, Projectable, Sortable {
```
aaron-steinfeld, line 11: Can bring over the comment about IdentifyingExpression or ColumnExpression here.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/RelationalExpression.java
```diff
@@ -0,0 +1,22 @@
...
+ *     company IN ('Traceable', 'Harness')
+ *  </code> can be constructed as <code>
+ *     RelationalExpression.of(
+ *         LiteralExpression.of("company"),
+ *         RelationalOperator.IN,
+ *         ConstantExpression.ofStrings("Traceable", "Harness"))));
+ * </code>
+ */
+@Value(staticConstructor = "of")
+public class RelationalExpression implements Filterable {
```
aaron-steinfeld, line 18: would `company EXISTS` be a relational expression? If we don't need it, can omit - but we should ensure that adding it would not change the design.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/RelationalOperator.java
```diff
@@ -0,0 +1,16 @@
...
+public enum RelationalOperator {
+  EQ,
+  NEQ,
+  GT,
+  LT,
+  GTE,
+  LTE,
+  IN,
+  CONTAINS,
+  EXISTS,
```
aaron-steinfeld, line 12: It would be :) so above comment applies.

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericOrderBy.java
```diff
@@ -0,0 +1,14 @@
...
+
+import lombok.Value;
+import org.hypertrace.core.documentstore.expression.Sortable;
+
+/**
+ * A generic ORDER BY definition that supports expressions. Note that this class is a more general
+ * version of {@link org.hypertrace.core.documentstore.OrderBy}
+ */
+@Value(staticConstructor = "of")
+public class GenericOrderBy {
```
aaron-steinfeld, line 11: The use of the term "Generic" here seems strange - isn't everything in this tree generic? If the idea is to keep it distinct from that other class, I think the package name would be the appropriate place to make that distinction (and a corresponding comment similar to what's here).

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericOrderBy.java
```diff
@@ -0,0 +1,14 @@
...
+  Sortable sortable;
+  boolean ascending;
```
aaron-steinfeld, line 13: personal nit: I've always found using a boolean to represent this weird, but may just be me. I just find something like `setDirection(DESC)` compared to `setAscending(false)` to be more readable.

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
```diff
@@ -0,0 +1,95 @@
...
+ *                 false))
+ *         .offset(5)
+ *         .limit(10)
+ *         .build();
+ *  </code>
+ */
+@Value
+@Builder
+public class GenericQuery {
+  List<Projection> projections;
```
aaron-steinfeld, line 63: naming: is selections more generic?

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
```diff
@@ -0,0 +1,95 @@
...
+  Filterable whereFilter;
```
aaron-steinfeld, line 64: we should try to take as much SQL lingo out as we can without losing clarity, given this is an abstraction over some fairly different styles of database.

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Singular List<Groupable> groupBys;
+  Filterable havingFilter;
+
+  @Singular List<GenericOrderBy> orderBys;
+
+  int offset;
```
aaron-steinfeld, line 71: I don't know what our existing support looks like - do we normally work in cursors, offsets or both?

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
```diff
@@ -0,0 +1,95 @@
...
+  int limit;
+
+  public static class GenericQueryBuilder {
```
aaron-steinfeld, line 74: Why do we need our own builder - are there shortcomings in what we can represent in lombok?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/AggregateExpression.java
```diff
@@ -0,0 +1,16 @@
...
+  Aggregator aggregator;
+  Projectable expression;
```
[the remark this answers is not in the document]
suresh-prakash: It would have to be translated to
`SELECT COUNT(1)`, which can then be substituted with `ConstantExpression.of(1)`

document-store/src/main/java/org/hypertrace/core/documentstore/expression/AggregateExpression.java
suresh-prakash: This is very useful! Thank you. :) 
Read this article https://hackernoon.com/are-interfaces-code-smell-bd19abc266d3 and changed it to a visitor design pattern. The only difference here is that we have multiple hierarchies here instead of one.


On another note, ConstantExpressions are projectable and not sortable though SQL allows integers in "ORDER BY" clause. For example, ORDER BY 2, sorts by 2nd field in the projection list. We do not support that as of today as this implementation is not going to be tied to any specific SQL.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/ArithmeticExpression.java
suresh-prakash: Renamed it as FunctionExpression as "SUBSTRING", "ADD", "SUBTRACT" are all functions. Would be happy to change it if anyone has a better alternative.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Filterable.java
suresh-prakash: While that's true, MongoDB allows alias names to be used in "HAVING" clauses. So, I think, it's better to leave it this way since an invalid expression would anyway fail while querying the DB.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Filterable.java
suresh-prakash: Makes sense! Thanks for the naming suggestions. :)

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericOrderBy.java
suresh-prakash: Makes sense. I also prefer this better.

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
suresh-prakash: Yup. Done some changes related to naming. Let me know if you find anything else specific to SQL.

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
suresh-prakash: I see that the existing implementation exposes offsets outside (and works with cursors internally).

document-store/src/main/java/org/hypertrace/core/documentstore/query/GenericQuery.java
suresh-prakash: This was done to avoid an unnecessary step while building the query. The implemented builder methods do the conversion from SelectingExpression to Selection internally, so that the clients can avoid the conversion step every single time.
`query.selection(Selection.of(LiteralExpression.of("col1")))` can be simplified as `query.selection(LiteralExpression.of("col1"))`.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/Aggregator.java
varsha-abhinandan: +1

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoFilteringExpressionParser.java
```diff
@@ -0,0 +1,22 @@
...
+import org.hypertrace.core.documentstore.expression.LogicalExpression;
+import org.hypertrace.core.documentstore.expression.RelationalExpression;
+import org.hypertrace.core.documentstore.parser.IFilteringExpressionParser;
+
+public class MongoFilteringExpressionParser implements IFilteringExpressionParser {
+  private final static String FILTER_CLAUSE = "$match";
+
+  @Override
+  public Object parse(LogicalExpression expression) {
+    return null;
```
suresh-prakash, line 14: Have just implemented a sample section. Will implement the rest once the API is accepted.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoQueryParserTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.hypertrace.core.documentstore.expression.ConstantExpression;
+import org.hypertrace.core.documentstore.expression.FilteringExpression;
+import org.hypertrace.core.documentstore.expression.IdentifierExpression;
+import org.hypertrace.core.documentstore.expression.RelationalExpression;
+import org.hypertrace.core.documentstore.expression.RelationalOperator;
+import org.hypertrace.core.documentstore.query.Query;
+
+// TODO: Remove this testing class
+public class MongoQueryParserTest {
```
suresh-prakash, line 11: Test class. Will be removed after necessary UTs are added.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoRelationalExpressionParser.java
```diff
@@ -0,0 +1,76 @@
...
+    SelectingExpression expression2 = expression.getOperand2();
+
+    MongoSelectingExpressionParser selectionParser = new MongoSelectingExpressionParser();
+    String key = Objects.toString(expression1.parse(selectionParser));
+    Object value = expression2.parse(selectionParser);
+
+    return generateMap(key, value, operator);
+  }
+
+  private static Map<String, Object> generateMap(String key, Object value, RelationalOperator operator) {
```
suresh-prakash, line 26: This method is copied from the existing implementation

document-store/src/main/java/org/hypertrace/core/documentstore/parser/IFilteringExpressionParser.java
```diff
@@ -0,0 +1,10 @@
+package org.hypertrace.core.documentstore.parser;
+
+import org.hypertrace.core.documentstore.expression.LogicalExpression;
+import org.hypertrace.core.documentstore.expression.RelationalExpression;
+
+public interface IFilteringExpressionParser {
```
varsha-abhinandan, line 6: It's a personal preference but I prefer not using the prefix "I" for interfaces since interfaces are what gets used widely and it hurts readability. We could just simply use FilteringExpressionParser since the impl class anyway has the db type prefix like MongoFilteringExpressionParser

document-store/src/main/java/org/hypertrace/core/documentstore/expression/AggregationOperator.java
```diff
@@ -0,0 +1,8 @@
+package org.hypertrace.core.documentstore.expression;
+
+public enum AggregationOperator {
```
varsha-abhinandan, line 3: can we move all the various operators under a package of its own and similarly the interfaces and the impl classes

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -0,0 +1,93 @@
...
+ *                     .operand(IdentifierExpression.of("col8")).build(),
+ *                 DESC))
+ *         .paginationDefinition(PaginationDefinition.of(5, 10))
+ *         .build();
+ *  </code>
+ */
+@Value
+@Builder
+public class Query {
+  List<Selection> selections;
```
varsha-abhinandan, line 62: Can't we use SelectionExpression here similar to FilteringExpression , GroupingExpression?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/FilteringExpression.java
```diff
@@ -0,0 +1,11 @@
+package org.hypertrace.core.documentstore.expression;
+
+import org.hypertrace.core.documentstore.parser.FilteringExpressionParser;
+
+/**
+ * An interface to represent that the expression can be used in either the WHERE clause or the
```
varsha-abhinandan, line 6: It might be cleaner to separate out the where and having expressions. Having clause is for aggregates. So, separating these two will make it less error-prone

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
suresh-prakash: Selection is a combination of SelectionExpression and an alias. So, SelectionExpression can't be directly used here.

varsha-abhinandan: If an alias is required for all selections, should we have this in the abstract class and have the impls of  SelectionExpression extend it?

suresh-prakash: Even I thought of doing the same. However, there are 2 reasons why it is not done that way.

1. Not all subclasses of SelectingExpression will need an "alias" field, because a subclass of SelectingExpression might actually be used as a SortingExpression or vice versa. Here, inheritance is purely used for type-substitution. Introducing "alias" to SelectingExpression as a class will end up in not only unwanted memory, but, also change the way a Query is constructed (in an undesired manner).
2. Similar to SelectingExpression, SortingExpression can also be converted to an abstract class with the "SortingOrder" field. When we do so, classes like "AggregateExpression" will need multiple inheritance, which is not supported in Java.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/FilteringExpression.java
suresh-prakash: Even Aaron had given the same comment. Ref.: https://github.com/hypertrace/document-store/pull/58#discussion_r745134184.

1. Even after separating out HavingExpression separately, it would have an identical structure since databases like MongoDB support aliases (IdentifierExpression) to be used in the conditional statements (than just AggregateExpressions)
2. For databases with different HAVING and WHERE implementations (like Postgres), it can be separated out at the Parser hierarchy.

In summary, a filter is common despite its usage in WHERE/HAVING clauses. The differentiation can be done in the implementation (at the Parser hierarchy).

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoFunctionExpressionParser.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.stream.Collectors;
+import org.hypertrace.core.documentstore.expression.impl.FunctionExpression;
+import org.hypertrace.core.documentstore.parser.SelectingExpressionParser;
+
+public class MongoFunctionExpressionParser {
+
+  static Map<String, Object> parse(FunctionExpression expression) {
+    int numArgs = expression.getOperands().size();
+
+    if (numArgs == 0) {
```
suresh-prakash, line 14: Probably these validations are no longer required with Hibernate validations taking place while building the query.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hypertrace.core.documentstore.expression.impl.ConstantExpression;
+import org.hypertrace.core.documentstore.expression.impl.FunctionExpression;
+import org.hypertrace.core.documentstore.expression.impl.IdentifierExpression;
+import org.hypertrace.core.documentstore.parser.SelectingExpressionParser;
+
+public class MongoSelectingExpressionParser implements SelectingExpressionParser {
+
+  @Override
+  public Object parse(AggregateExpression expression) {
+    return null;
```
suresh-prakash, line 13: TODO

saxenakshitiz:
Overall lgtm

document-store/src/main/java/org/hypertrace/core/documentstore/query/Selection.java
```diff
@@ -0,0 +1,35 @@
...
+ * {@link IdentifierExpression}, the alias is inferred if not supplied.
+ */
+@Value
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class Selection {
+
+  // Special selection representing the selection of all the columns
+  public static final Selection ALL = new Selection(null, null);
+
+  @NotNull SelectingExpression expression; // Can be null only in the case of select all columns
```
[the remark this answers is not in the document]
varsha-abhinandan: As discussed, let's model ALL as well as a selection expression instead of dealing with nulls.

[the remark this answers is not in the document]
suresh-prakash: Sure

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -0,0 +1,138 @@
...
+  @Size(min = 1)
+  List<@NotNull SortingDefinition> sortingDefinitions;
+
+  // Missing pagination definition represents fetching all the records
+  PaginationDefinition paginationDefinition;
+
+  public static class QueryBuilder {
+
+    public Query build() {
+      fillMissingSelections();
```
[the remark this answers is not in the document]
suresh-prakash: Yeah. That makes perfect sense.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoAggregateExpressionParser.java
```diff
@@ -0,0 +1,17 @@
+package org.hypertrace.core.documentstore.mongo.parser;
+
+import java.util.Map;
+import org.hypertrace.core.documentstore.expression.impl.AggregateExpression;
+import org.hypertrace.core.documentstore.parser.SelectingExpressionParser;
+
+public class MongoAggregateExpressionParser {
+
+  static Map<String, Object> parse(final AggregateExpression expression) {
+    String key = "$" + expression.getAggregator().name().toLowerCase();
```
varsha-abhinandan, line 10: nit: can we extract this out to a contant

varsha-abhinandan:
@suresh-prakash Rest of the skeletal LGTM

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
```diff
@@ -0,0 +1,8 @@
+package org.hypertrace.core.documentstore.expression.operators;
+
+public enum AggregationOperator {
```
varsha-abhinandan, line 3: min, max

varsha-abhinandan: though may not be needed currently

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -0,0 +1,135 @@
...
+
+  FilteringExpression filter;
+
+  @Singular
+  @Size(min = 1)
+  List<@NotNull GroupingExpression> aggregations;
+
+  FilteringExpression aggregationFilter;
+
+  @Size(min = 1)
```
kotharironak, line 84: Does this mean order by being mandatory in a request or will have a default?

document-store/src/main/java/org/hypertrace/core/documentstore/query/WhitelistedSelection.java
```diff
@@ -0,0 +1,42 @@
...
+    return expression instanceof AggregateExpression;
+  }
+
+  @Override
+  public boolean allColumnsSelected() {
+    return false;
+  }
+
+  public static Selection of(final SelectingExpression expression) {
+    return WhitelistedSelection.of(expression, null);
```
kotharironak, line 36: why does validation not required in this method? Is validation is only for `alias`?

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -0,0 +1,135 @@
...
+@Builder
+public class Query {
+
+  @NotEmpty List<@NotNull Selection> selections;
+
+  FilteringExpression filter;
+
+  @Singular
+  @Size(min = 1)
+  List<@NotNull GroupingExpression> aggregations;
```
kotharironak, line 80: I see that for `Selection` we have `definition and `Expression` objects separate. But, why are we not having the same pattern for `GroupBy` and `Filter`?

document-store/src/main/java/org/hypertrace/core/documentstore/parser/SelectingExpressionParser.java
```diff
@@ -0,0 +1,16 @@
+package org.hypertrace.core.documentstore.parser;
+
+import org.hypertrace.core.documentstore.expression.impl.AggregateExpression;
+import org.hypertrace.core.documentstore.expression.impl.ConstantExpression;
+import org.hypertrace.core.documentstore.expression.impl.FunctionExpression;
+import org.hypertrace.core.documentstore.expression.impl.IdentifierExpression;
+
+public interface SelectingExpressionParser {
+  Object parse(AggregateExpression expression);
```
kotharironak, line 9: Like `SortingExpressionParser`, shouldn't the rest of the interfaces of parse be taking `final` argument?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -0,0 +1,99 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.bson.conversions.Bson;
+import org.hypertrace.core.documentstore.query.PaginationDefinition;
+import org.hypertrace.core.documentstore.query.Query;
+
+@Slf4j
+public class MongoQueryExecutor {
+
+  public static Iterator<BasicDBObject> find(
+      final Query query, final com.mongodb.client.MongoCollection<BasicDBObject> collection) {
```
kotharironak, line 32: Can we make these methods `non-static` and `collection will be its instance as mostly they will be used from the implementation of `Collection` interface?

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
suresh-prakash: Actually, it meant this value can be NULL or a list with at least one element (but, never an empty list).
However, after going through this in our discussion, even having an empty list will not harm as it is handled.
So, will remove this annotation from here.

document-store/src/main/java/org/hypertrace/core/documentstore/query/WhitelistedSelection.java
suresh-prakash: Validation is required here. The invoked method `WhitelistedSelection.of()` does the validation internally.
Validation is only for the SelectingExpression. "alias" doesn't require any validation as it can be null/blank.

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
suresh-prakash: Yeah. I see the issue in readability. Will consider this to improve readability in subsequent PRs (along with writing integration tests).

suresh-prakash: Keeping this conversation unresolved to address in a future PR.

document-store/src/main/java/org/hypertrace/core/documentstore/parser/SelectingExpressionParser.java
suresh-prakash: Yup. Good catch. I added the missing `final`s

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/RelationalExpression.java
```diff
@@ -0,0 +1,45 @@
...
+ * </code> can be constructed as <code>
+ * RelationalExpression.of( IdentifierExpression.of("company"), RelationalOperator.IN,
+ * ConstantExpression.ofStrings("Traceable", "Harness"))));
+ * </code>
+ */
+@Value
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class RelationalExpression implements FilteringExpression {
+
+  @NotNull SelectingExpression operand1;
```
kotharironak, line 28: nit: I think, operand1 to lhs and operand2 to rhs would have been more readable

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -0,0 +1,104 @@
...
+    BasicDBObject skipClause = getSkipClause(query.getPaginationDefinition());
+    BasicDBObject limitClause = getLimitClause(query.getPaginationDefinition());
+
+    BasicDBObject projectClause = getProjectClause(query.getSelections());
+
+    List<BasicDBObject> pipeline =
+        Stream.of(
+                filterClause,
+                groupClause,
+                groupFilterClause,
```
kotharironak, line 70: Order will not matter, right?
Should `groupFilterClause` come before `groupClause`?

suresh-prakash: Yes. The order of specification is immaterial

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/RelationalExpression.java
suresh-prakash: Makes sense. Will address in the follow-up PR

kotharironak (approved):
Overall LGTM. I will try out a few things once we have an integration test to play with as follow-up PR. See if any open comments from others.

document-store/src/main/java/org/hypertrace/core/documentstore/query/AllSelection.java
```diff
@@ -0,0 +1,20 @@
+package org.hypertrace.core.documentstore.query;
+
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public class AllSelection implements Selection {
```
varsha-abhinandan, line 7: Thinking about this again - we probably don't need to support "all". Expecting specific fields should be good enough.

suresh-prakash: Yeah. With respect to MongoDB, selecting specific-fields is sufficient. Since Postgres (which is an SQL) is also supported as of now and for any future use cases, I think, it is good to have.

Since these APIs are not bound to a specific database, I wanted to keep it as generic as possible so that we don't break the compatibility frequently.

What do you think?

varsha-abhinandan: Yeah, this isn't specific to a particular DB. It's whether our query interface needs to support "all" fields or not. The existing query interface also expects the user to specify the required fields upfront. In the new Query class, the top fields seem to have a mix of expressions, definitions, and now the selection interface. Was thinking if we can at least simplify the selection one by removing the * support and look into the other two. We can merge this PR and take the open comments from Ronak as well in your next.

varsha-abhinandan (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/query/AllSelection.java
suresh-prakash: Sure. Will work on it in the next PR.

PR #60, opened 2021-11-19
Query transformation support for MongoDB

suresh-prakash:
## Description
* Introduced the query transformation
* Implemented DISTINCT_COUNT() using the transformation function
* Introduced the different types of MongoSelectingExpressionParser class with decorator design pattern
* Refactored MongoRelationalExpressionParser class with currying to avoid code duplication
* Unit & integration tests for the new functionality
* A few bug fixes

### Testing
* Added unit test
* Added integration test

### Checklist:

### Documentation


document-store/src/main/java/org/hypertrace/core/documentstore/query/SortingSpec.java
```diff
@@ -15,6 +17,7 @@
  */
 @Value
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode(cacheStrategy = CacheStrategy.LAZY)
```
varsha-abhinandan, line 20: query: what's the benefit of this since the fields are not final?

document-store/build.gradle.kts
```diff
@@ -24,6 +24,7 @@ dependencies {
   implementation("com.google.guava:guava-annotations:r03")
   implementation("org.apache.commons:commons-lang3:3.10")
   implementation("net.jodah:failsafe:2.4.0")
+  implementation("com.google.guava:guava:16.0.1")
```
varsha-abhinandan, line 27: Can we use the latest version?

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -93,7 +99,7 @@ public static QueryBuilder builder() {
       return Collections.emptyList();
     }
 
-    return aggregation.getExpressions();
+    return Collections.unmodifiableList(aggregation.getExpressions());
```
varsha-abhinandan, line 102: Just thinking when would the clients use these getters which returns unmodifiableList ?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoAggregationSelectingExpressionParser.java
```diff
@@ -0,0 +1,35 @@
...
+    super(query);
+  }
+
+  @Override
+  public Map<String, Object> visit(AggregateExpression expression) {
+    return new MongoAggregateExpressionParser(query).parse(expression);
+  }
+
+  @Override
+  public Object visit(final ConstantExpression expression) {
```
varsha-abhinandan, line 22: Couldn't understand the reason for this change. I might need a code walkthrough with regards to these parser changes.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsAddingTransformation.java
```diff
@@ -0,0 +1,107 @@
...
+  @SuppressWarnings("unchecked")
+  @Override
+  public Optional<SelectionSpec> visit(AggregateExpression expression) {
+    if (expression.getAggregator() == DISTINCT_COUNT) {
+      // Since MongoDB doesn't support $distinctCount in aggregations, we convert this to
+      // $addToSet function. So, we need to project $size(set) instead of just the alias
+      SelectingExpression pairingExpression =
+          FunctionExpression.builder()
+              .operator(LENGTH)
+              .operand(IdentifierExpression.of(alias))
```
varsha-abhinandan, line 82: Could the alias be null/empty here since it's optional while creating the SelectionSpec ?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsAddingTransformer.java
```diff
@@ -0,0 +1,30 @@
...
+import org.hypertrace.core.documentstore.query.QueryTransformer;
+import org.hypertrace.core.documentstore.query.SelectionSpec;
+
+final class MongoSelectionsAddingTransformer implements QueryTransformer {
+  @Override
+  public Query transform(final Query query) {
+    List<SelectionSpec> newSpecs = new ArrayList<>();
+
+    for (SelectionSpec spec : query.getSelections()) {
+      String alias = spec.getAlias();
```
varsha-abhinandan, line 18: Same question regarding alias being null/empty

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsUpdatingTransformation.java
```diff
@@ -0,0 +1,143 @@
...
+  @SuppressWarnings("unchecked")
+  @Override
+  public SelectionSpec visit(final IdentifierExpression expression) {
+    GroupingExpression matchingGroup = aggregationMap.get(expression.hashCode());
+    if (!expression.equals(matchingGroup)) {
+      return source;
+    }
+
+    String key = expression.getName();
+    String identifier = MongoCollection.ID_KEY + "." + key;
```
varsha-abhinandan, line 129: nit : extract into constant

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
```diff
@@ -278,6 +277,26 @@ public void testAggregateWithNestedFields() throws IOException {
...
 
+  @Test
+  public void testDistinctCount() throws IOException {
+    Query query =
+        Query.builder()
+            .addSelection(
+                AggregateExpression.of(DISTINCT_COUNT, IdentifierExpression.of("quantity")),
+                "qty_count")
+            .addSelection(IdentifierExpression.of("item"))
+            .addAggregation(IdentifierExpression.of("item"))
```
varsha-abhinandan, line 288: Can we also test with multiple group by columns and without an alias?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsUpdatingTransformation.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private SelectionSpec substitute(final AggregateExpression expression) {
+    return Optional.ofNullable(AGGREGATION_SUBSTITUTE_MAP.get(expression.getAggregator()))
+        .map(newExp -> SelectionSpec.of(newExp.apply(expression), source.getAlias()))
+        .orElse(source);
+  }
+
+  private Map<Integer, GroupingExpression> getAggregationMap() {
+    return aggregations.stream().collect(toMap(GroupingExpression::hashCode, identity()));
```
varsha-abhinandan, line 141: why do we need this map whose key is a hashcode?  How are we ensuring these grouping expressions don't get modified/transformed in the future? Seems error-prone.

suresh-prakash: > why do we need this map whose key is a hashcode?
This is to identify a matching GroupingExpression in constant time. Without this, we will have to iterate through the list, which make this transformation take linear time.

> How are we ensuring these grouping expressions don't get modified/transformed in the future?
All the implementations of GroupingExpression are `@Value` classes (they are immutable). Hence, their hash code's can't change (during the lifecycle of the objects).

> Seems error-prone.
We are updating only the SelectingExpressions (or SelectionSpecs). None of the GroupingExpression are replaced/transformed in the query.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
suresh-prakash: Will do.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsUpdatingTransformation.java
suresh-prakash: You mean, make identifier as a final variable?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsAddingTransformation.java
suresh-prakash: Yeah. Good catch. Need to make alias mandatory for AggregationExpression since we cannot work without them in MongoDB. Will handle it.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsAddingTransformer.java
suresh-prakash: Yeah. Good catch. Need to make alias mandatory for AggregationExpression since we cannot work without them in MongoDB. Will handle it.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoAggregationSelectingExpressionParser.java
suresh-prakash: Sure. I will also add some code comments to make it verbose.

In summary, we have 3 major implementations of MongoSelectingExpressionParsers:

1. MongoSelectingExpressionParser: Parses all types of expressions
2. MongoAggregationSelectingExpressionParser: Parse only AggregationExpressions (since they should go in the "$group" clause)
3. MongoNonAggregationSelectingExpressionParser: Parse all other expression types (other than AggregationExpression) (since they should go in the "$project" clause)

On top of these, we have 2 decorators, which may perform some additional operations on the parsed response.
(a). MongoIdentifierPrefixingSelectingExpressionParser: Prefixes IdentifierExpressions with "$" in front. 
(b). MongoProjectionSelectingExpressionParser: Wraps the parsed responses with supplied/generated aliases.

We use any combinations of decorators (a) and (b) with any one (1), (2) and (3) at different places depending upon the context.

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
suresh-prakash: The client's wouldn't have a case to use these. These methods would mostly be used by the parsers.

document-store/build.gradle.kts
suresh-prakash: Sure. Will do.

document-store/src/main/java/org/hypertrace/core/documentstore/query/SortingSpec.java
suresh-prakash: All the fields and the class are implicitly final through the `@Value` annotation above.
Since this is basically an immutable class, we cache the hashCode for optimization.

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -105,13 +111,50 @@ public static QueryBuilder builder() {
...
-    return sort.getSortingSpecs();
+    return Collections.unmodifiableList(sort.getSortingSpecs());
   }
 
   public Optional<Pagination> getPagination() {
     return Optional.ofNullable(pagination);
   }
 
+  @AllArgsConstructor
+  public static final class TransformedQueryBuilder {
```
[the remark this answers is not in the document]
suresh-prakash: Good point. I was in dilemma as well. If we take this class outside this package, we may have to make the Query constructor public. Moreover, we may need to provide getters for all the fields.

So, I thought, it's better to keep this "friend" class here so that any query construction cannot happen outside of the Query class. Considering this as a special type of builder of the Query class, so that we do not have to expose the Query constructor outside (since we also need to perform the hibernate validations on top).

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsUpdatingTransformation.java
suresh-prakash: I have renamed "aggregations" to "groupingExpressions" for improved clarity.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
```diff
@@ -1,74 +1,100 @@
...
+public abstract class MongoSelectingExpressionParser implements SelectingExpressionVisitor {
 
   private static final String PROJECT_CLAUSE = "$project";
 
-  private final String identifierPrefix;
+  protected final MongoSelectingExpressionParser baseParser;
 
-  public MongoSelectingExpressionParser(final Query query) {
-    this(query, false);
+  protected MongoSelectingExpressionParser() {
```
varsha-abhinandan, line 23: query: where is this constructor used?

suresh-prakash: In some of the sub-classes. Ref. image below.
image

document-store/src/main/java/org/hypertrace/core/documentstore/query/SortingSpec.java
suresh-prakash: As discussed, I'm removing these.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoAggregationSelectingExpressionParser.java
suresh-prakash: Refactored to improve flexibility, reliability and readability.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsUpdatingTransformation.java
suresh-prakash: As discussed, got rid of the map by foregoing optimization for better readability and robustness.

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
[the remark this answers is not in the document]
varsha-abhinandan: Still not comfortable exposing this internals to the clients interacting with the Query class. 
Prefer extracting this out to a separate class within the query transformer package. Instead of making the query constructor public, I believe we can internally use the Query builder again. We anyways have exposed a getter for all the fields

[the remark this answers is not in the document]
suresh-prakash: Makes sense. I have moved this class to a separate query transformer package (outside of the mongo package).

varsha-abhinandan (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
```diff
@@ -3,6 +3,8 @@
 public enum AggregationOperator {
   AVG,
   COUNT,
+  DISTINCT,
```
kotharironak, line 6: what difference between `DISTINCT` and `DISTINCT_Count`?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/FunctionOperator.java
```diff
@@ -4,6 +4,7 @@
   // Unary operations
   ABS,
   FLOOR,
+  LENGTH,
```
kotharironak, line 7: I hope, all these operations are in need, right?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoAggregateExpressionParser.java
```diff
@@ -1,25 +1,59 @@
...
+import org.hypertrace.core.documentstore.parser.SelectingExpressionVisitor;
+
+@NoArgsConstructor
+final class MongoAggregateExpressionParser extends MongoSelectingExpressionParser {
+  private static final Map<AggregationOperator, String> KEY_MAP =
+      unmodifiableMap(
+          new EnumMap<>(AggregationOperator.class) {
+            {
+              put(AVG, "$avg");
+              put(DISTINCT, "$addToSet");
```
kotharironak, line 25: `DISTINCT_COUNT` is not required?

suresh-prakash: Mongo doesn't have a support for DISTINCT_COUNT.

It is internally converted to DISTINCT by the query transformers.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/FunctionOperator.java
suresh-prakash: As of now, we use only LENGTH. 

However, would like to keep minimalistic operations here to avoid back-and-forth library changes.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
suresh-prakash: `DISTINCT` will fetch the distinct values. `DISTINCT_COUNT` will get the number of distinct values.
In other words, `DISTINCT` returns a Set where `DISTINCT_COUNT` returns the size of that set.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoLogicalExpressionMongoParser.java
```diff
@@ -0,0 +1,42 @@
...
+    FilteringExpressionVisitor parser = new MongoFilteringExpressionParser();
+    List<Object> parsed =
+        expression.getOperands().stream()
+            .map(exp -> exp.visit(parser))
+            .collect(Collectors.toList());
+
+    LogicalOperator operator = expression.getOperator();
+    String key = KEY_MAP.get(operator);
+
+    if (key == null) {
```
kotharironak, line 36: Shouldn't the operator check be at the beginning? So, we can avoid parsing on un-supported operators.

I think this will suffice, right?
`KEY_MAP.containsKey(operator)`?

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -61,227 +48,8 @@
...
-  private final Selection selection; // Missing selection represents fetching all the columns
-  private final Filter filter;
-
-  private final Aggregation aggregation;
-  private final Filter aggregationFilter;
-
-  private final Sort sort;
-  private final Pagination pagination; // Missing pagination represents fetching all the records
-
+public abstract class Query {
```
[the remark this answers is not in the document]
suresh-prakash: This was basically done to abstract out the internals of the Query class.
The clients should know nothing about the Query class. They should be able to simply build it and pass it on to `find` or `aggregate`.

Essentially, the new design enables us to change the QueryInternal structure the way we want.
Suppose, we have a new AggregationOperator (not exposed outside) that needs to be applied only by the Query Transformer, then, it's easy to make these changes in the QueryInternal class.
Basically, we are making the Query class an abstraction instead of an implementation. This way, the clients are agnostic of how the query class is used internally. For example, we may change the QueryInternal class to be mutable tomorrow (instead of building a whole new QueryInternal class) without affecting the client code. This also goes by the design principle that says "code for abstraction".

We don't have anything other than QueryInternal in mind as of now. But, this gives us a flexibility to do so if required in future.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -16,25 +16,27 @@
...
+import org.hypertrace.core.documentstore.query.QueryInternal;
 
 @Slf4j
 @AllArgsConstructor
 public class MongoQueryExecutor {
   final com.mongodb.client.MongoCollection<BasicDBObject> collection;
 
-  public MongoCursor<BasicDBObject> find(final Query query) {
+  public MongoCursor<BasicDBObject> find(final Query originalQuery) {
+    QueryInternal query = (QueryInternal) originalQuery;
```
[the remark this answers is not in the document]
suresh-prakash: >If, we are going to cast this, why do we need QueryInternal?
Answered here: https://github.com/hypertrace/document-store/pull/60#discussion_r759220385

> if the originalQuery is not an instance of QueryInternal, would it throw an exception?
As of today, since we do not have any other implementation, this is fine. But, when we have more than one sub-class of Query, the idea is to introduce a factory or so here. However, we anticipate that there won't be a need for any other sub-class of the Query class.

document-store/src/main/java/org/hypertrace/core/documentstore/query/QueryInternal.java
```diff
@@ -0,0 +1,61 @@
...
+  private final Filter filter;
+
+  private final Aggregation aggregation;
+  private final Filter aggregationFilter;
+
+  private final Sort sort;
+  private final Pagination pagination; // Missing pagination represents fetching all the records
+
+  public List<SelectionSpec> getSelections() {
+    if (selection == null) {
```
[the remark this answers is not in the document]
suresh-prakash: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoLogicalExpressionMongoParser.java
suresh-prakash: > Shouldn't the operator check be at the beginning?

Done.

> I think this will suffice, right? `KEY_MAP.containsKey(operator)`?
Using `containsKey` would still require us another lookup as we are using the value (in the return statement).

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -56,8 +57,7 @@
 
   /**
    * Search for documents matching the query. For advanced functionalities like selecting summing
-   * column values, sorting by expressions, etc., ref. {@link
-   * #find(org.hypertrace.core.documentstore.query.Query)}
+   * column values, sorting by expressions, etc., ref. {@link #find(QueryInternal)}
```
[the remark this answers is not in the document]
suresh-prakash: Good catch. Fixed.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoIdentifierPrefixingSelectingExpressionParser.java
```diff
@@ -0,0 +1,22 @@
...
+
+  MongoIdentifierPrefixingSelectingExpressionParser(
+      final MongoSelectingExpressionParser baseParser) {
+    super(baseParser);
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public String visit(final IdentifierExpression expression) {
+    assert baseParser != null;
```
kotharironak, line 19: In the parent class, if `baseParse` is null, we are throwing `throw getUnsupportedOperationException(expression)`, and here we are asserting, what's the difference?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoProjectionSelectingExpressionParser.java
```diff
@@ -0,0 +1,95 @@
...
+  MongoProjectionSelectingExpressionParser(
+      final String alias, final MongoSelectingExpressionParser baseParser) {
+    super(baseParser);
+    this.alias = alias;
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public Map<String, Object> visit(final AggregateExpression expression) {
+    assert baseParser != null;
```
kotharironak, line 27: why would the `basePraser` be null? Is it possible to avoid that?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
```diff
@@ -1,74 +1,99 @@
...
+    this(null);
```
kotharironak, line 24: Is it possible to create a default no-op selection parser?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSortingExpressionParser.java
```diff
@@ -9,51 +15,67 @@
...
+  @SuppressWarnings("unchecked")
   @Override
-  public Map<String, Object> parse(final IdentifierExpression expression) {
-    String parsed = new MongoIdentifierExpressionParser(query).parse(expression);
-    return Map.of(parsed, getOrder());
+  public Map<String, Object> visit(final IdentifierExpression expression) {
+    String parsed = new MongoIdentifierExpressionParser().parse(expression);
+    Integer value = ORDER_MAP.get(order);
+
+    if (value == null) {
```
kotharironak, line 66: I think we can check this beginning of the method before parsing. If the operator is not supported, we fail fast.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoIdentifierPrefixingSelectingExpressionParser.java
suresh-prakash: Got rid with NoOpParser.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoProjectionSelectingExpressionParser.java
suresh-prakash: Got rid with NoOpParser.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
suresh-prakash: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSortingExpressionParser.java
suresh-prakash: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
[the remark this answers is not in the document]
kotharironak: As discussed offline would be better to go with a concrete Query interface for method (find and aggregate)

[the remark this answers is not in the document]
suresh-prakash: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -82,20 +82,32 @@
...
-  private void logPipeline(List<BasicDBObject> pipeline) {
+  private void logPipeline(final List<BasicDBObject> pipeline) {
     log.debug(
         "MongoDB aggregation():\n Collection: {}\n Pipeline: {}",
         collection.getNamespace(),
         pipeline);
   }
+
+  private Query transformAndLog(Query query) {
+    log.debug("MongoDB query before transformation: {}", query);
```
kotharironak, line 108: nit: as follow up, log only on exception or something. Otherwise, this will flood the number of logs or rate limit them.

kotharironak (approved):
lgtm, a large one!!

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
suresh-prakash: Sure

PR #64, opened 2021-12-16
Modified hibernate-validator code and dependency to avoid exception in consumers.

singhalprerana:
Having this module as dependency in callers sometimes results in `javax.validation.ValidationException: HV000183`
Modified the code and the dependencies in an attempt to avoid this


singhalprerana:
Closing in favour of https://github.com/hypertrace/document-store/pull/65

PR #66, opened 2021-12-30
Handling duplicate selections in MongoDB

suresh-prakash:
Handle duplications using
(i) a overwrite strategy for duplicate selections
(ii) an error-out strategy for duplicate aliases

## Description
Duplicate selections were causing a problem. This PR is to fix the issue.

### Testing
Added necessary unit and integration tests

### Checklist:

### DIFF
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
@@ -128,6 +128,32 @@ public class MongoQueryExecutorIntegrationTest {
   }
 
   @Test
+  public void testFindWithDuplicateSelections() throws IOException {
+    List<SelectionSpec> selectionSpecs =
+        List.of(
+            SelectionSpec.of(IdentifierExpression.of("item")),
+            SelectionSpec.of(IdentifierExpression.of("item")),
+            SelectionSpec.of(IdentifierExpression.of("price")),
+            SelectionSpec.of(IdentifierExpression.of("quantity")),
+            SelectionSpec.of(IdentifierExpression.of("quantity")),
+            SelectionSpec.of(IdentifierExpression.of("date")));
+    Selection selection = Selection.builder().selectionSpecs(selectionSpecs).build();
+    Filter filter =
+        Filter.builder()
+            .expression(
+                RelationalExpression.of(
+                    IdentifierExpression.of("item"),
+                    NOT_IN,
+                    ConstantExpression.ofStrings(List.of("Soap", "Bottle"))))
+            .build();
+
+    Query query = Query.builder().setSelection(selection).setFilter(filter).build();
+
+    Iterator<Document> resultDocs = collection.find(query);
+    assertDocsEqual(resultDocs, "mongo/simple_filter_response.json");
+  }
+
+  @Test
   public void testFindWithSortingAndPagination() throws IOException {
     List<SelectionSpec> selectionSpecs =
         List.of(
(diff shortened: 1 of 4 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
```diff
@@ -79,4 +83,12 @@ public static BasicDBObject getProjectClause(final Query query) {
 
     return spec.getExpression().visit(parser);
   }
+
+  private static <T> T mergeValues(final T first, final T second) {
+    if (first.equals(second)) {
+      return second;
+    }
+
+    throw new IllegalArgumentException("Query contains duplicate aliases");
```
skjindal93, line 92: didn't understand the exception message. Won't duplicate selections have duplicate aliases?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
```diff
@@ -127,6 +127,32 @@ public void testFindSimple() throws IOException {
...
+                RelationalExpression.of(
+                    IdentifierExpression.of("item"),
+                    NOT_IN,
+                    ConstantExpression.ofStrings(List.of("Soap", "Bottle"))))
+            .build();
+
+    Query query = Query.builder().setSelection(selection).setFilter(filter).build();
+
+    Iterator<Document> resultDocs = collection.find(query);
+    assertDocsEqual(resultDocs, "mongo/simple_filter_response.json");
```
skjindal93, line 153: Shouldn't we be returning results for duplicate selections back, instead of de-duping it in the response? Basically, if a client asks for N number of selections, it expects N number of responses, corresponding to each selection.

Cc @avinashkolluru @aaron-steinfeld

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/parser/MongoSelectingExpressionParser.java
suresh-prakash: There can be 3 cases as follows.
1. Same selection and same alias as in `SELECT sum(price) AS total, sum(price) AS total`. In this case, both the keys (alias) and both the values (selection) match. Key: total, Value: sum(price). In this case, we pick the second value (equivalently we could have picked the first)
2. Same selection with different alias as in `SELECT sum(price) AS total, sum(price) AS price`. In this case, we don't have a problem since the keys (aliases) are unique and we wouldn't be hitting this method
3. Different selections with same aliases as in `SELECT SUM(quantity) AS total, SUM(price) AS total`. We shouldn't be allowing such cases as it clearly invites ambiguity. When the parsed values are different for the same key (alias), we detect it and throw exception.

suresh-prakash: Updating the message for clarity.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
suresh-prakash: Agree with you on this point. But, unfortunately, since Mongo returns results in the JSON format, I'm not sure how do we have duplicate keys in the same object.

A simple workaround for this is that the clients make sure to have different "aliases" for the same selection field, so that they are repeated in the response.

suresh-prakash: And not just Mongo, the method org.hypertrace.core.documentstore.Collection.aggregate() itself returns an iterator of org.hypertrace.core.documentstore.Document objects, which exposes only one method "toJson()".

avinashkolluru: In a traditional DB with columns/rows I agree with what you say @skjindal93 
But this is a document store interface. The response for the query is a Document. What does duplicate selections mean in a document(which is a JSON format). 
If you actually run a mongo query with duplicate projections the output contains it only once. In fact, I dont even think duplicates are possible in JSON.

avinashkolluru: Ahh... @suresh-prakash beat me to this. But, my thoughts are also aligned with what Suresh says.

aaron-steinfeld: Correct, dupe keys are not possible in JSON. We should error on dupe result keys (i.e. `alias.orElse(colName)`), support multiple identical selections with different result keys. Ran into this kind of thing a lot with attribute expression changes, we're _really_ bad about actually respecting the aliases provided.

skjindal93 (approved):

PR #73, opened 2022-03-01
Close mongodb cursor if the iterator has no more items

saxenakshitiz:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -582,7 +582,11 @@ private BasicDBObject selectionCriteriaForKeys(Set<Key> keys) {
     return new Iterator<>() {
       @Override
       public boolean hasNext() {
-        return cursor.hasNext();
+        boolean hasNext = cursor.hasNext();
+        if (!hasNext) {
+          cursor.close();
+        }
+        return hasNext;
```
avinashkolluru, line 589: How do we handle cases where next() throws an error. Wouldn't the cursor remain open in those cases

aaron-steinfeld: We should probably catch in that case. Would it be feasible to additionally make this a closable iterator? Basically what MongoCursor is doing. The whole design of doc store is problematic in that it returns an iterator which is left open until completely iterated - this fix only resolves the case where we complete iteration.

avinashkolluru: Yeah. A closeable iterator is a better option. But that would mean its the consumers responsibility to close now.

aaron-steinfeld: I was thinking both - closing it on completion like this PR is doing, but aso returning something that maintains compatibility with the existing API like
```java
interface Cursor<TResult> extends Iterator<TResult>, Closeable {}
```

saxenakshitiz: Made this change

document-store/src/main/java/org/hypertrace/core/documentstore/ClosableIterator.java
```diff
@@ -0,0 +1,6 @@
+package org.hypertrace.core.documentstore;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+public interface ClosableIterator<E> extends Iterator<E>, Closeable {}
```
aaron-steinfeld, line 6:
```suggestion
public interface CloseableIterator<E> extends Iterator<E>, Closeable {}
```
Should match jdk interface name

aaron-steinfeld (approved):

aaron-steinfeld (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/ClosableIterator.java
saxenakshitiz: Done

PR #76, opened 2022-03-07
Count query by filter

suresh-prakash:
## Description
Introduce counting documents based on the new filter DTO

### Testing
Added necessary integration test

### Checklist:

### Documentation

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
@@ -110,7 +110,7 @@ public interface Collection {
    * @return the number of documents matching the passed filter conditions Note that this method is
    *     a generic version of {@link #total(Query)}
    */
-  long count(final org.hypertrace.core.documentstore.query.Filter filter);
+  long total(final org.hypertrace.core.documentstore.query.Filter filter);
 
   /**
    * @return the total number of documents matching the query applying the filters passed, and
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
@@ -106,6 +106,11 @@ public interface Collection {
   /** @return the number of documents in the collection */
   long count();
 
+  /** @return the number of documents matching the passed filter conditions
+   * Note that this method is a generic version of {@link #count()}
+   */
+  long count(final org.hypertrace.core.documentstore.query.Filter filter);
+
   /**
    * @return the total number of documents matching the query applying the filters passed, and
    *     ignoring offset and limit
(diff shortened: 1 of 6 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -106,6 +106,12 @@
   /** @return the number of documents in the collection */
   long count();
 
+  /**
+   * @return the number of documents matching the passed filter conditions Note that this method is
+   *     a generic version of {@link #total(Query)}
+   */
+  long count(final org.hypertrace.core.documentstore.query.Filter filter);
```
skjindal93, line 113: Should it take the query object instead? How about the count, when there is a group by? Is it possible to call it total instead?

suresh-prakash: > Should it take the query object instead?

I don't think so. This method is clearly going to consume the filter today and forever. Unnecessary wrapping of filter inside the query object and then unwrapping in document-store seems pointless by creating a confusion to the clients about which fields would be used and which ones would not.

> How about the count, when there is a group by?

The `aggregate()` method should be relied on for such use-cases. Ref.: https://docs.mongodb.com/manual/reference/operator/aggregation/count/. They can simply execute a query similar to `SELECT COUNT(1) AS count FROM <collection> GROUP BY col1, col2, col3`.

Is it possible to call it total instead?

Sure. Done for consistency.

skjindal93: A bit confused. Is the expectation being the following?

- If the query has both filter and group by, we will be using `aggregate` method.
- If the query has only filter and group by, we will be using `total` method

suresh-prakash: That's correct.
Oh! Wait.
Actually, one can simply do `SELECT COUNT(1) AS count FROM <collection>` on `aggregate` without any group by. But, there are 2 issues with this.

1. The return value would be a Document, which needs to be parsed in order to get the desired count. (As opposed to getting a single `long` value)
2. We don't know the performance impact of performing `aggregate`, when we are looking for just the # of documents with some filters.

Now, I'm thinking, we should be relying on `aggregate` always (unless we see any performance issue). And, maybe provide a utility to extract count from Document object.

skjindal93: That was my concern too. 

We can still add a new total method, but it shouldn't just be taking filter. It can take the entire query object. And, let the implementation internally use `aggregate` if needed. 

And, extract count/total out from document

suresh-prakash: Just realised that `Counting by group` and `Counting groups` are 2 different things. What we want here is the latter.

This would be non-trivial to implement since we don't have the support for `COUNT` in the aggregation pipeline.

Closing this PR for now.

suresh-prakash:
Closing this PR for now.

PR #77, opened 2022-03-11
feat: bulk delete documents

skjindal93:


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -981,6 +982,23 @@ public void testDelete(String dataStoreName) throws IOException {
     Assertions.assertEquals(collection.count(), 0);
   }
 
+  @ParameterizedTest
+  @MethodSource("databaseContextProvider")
+  public void testBulkDelete(String dataStoreName) throws IOException {
```
kotharironak, line 987: Runs for both postgres and mongo, right?

skjindal93: yes

kotharironak (approved):

PR #79, opened 2022-04-08
Filter post unwind

rish691:
Effect of applying filter pre and post unwind is different, since unwind creates multiple document for the source document. 
Applying the same filters again.

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
@@ -37,7 +37,6 @@ public class MongoQueryExecutor {
           List.of(
               query -> singleton(getFilterClause(query, Query::getFilter)),
               MongoFromTypeExpressionParser::getFromClauses,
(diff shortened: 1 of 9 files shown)


varsha-abhinandan (dismissed):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -79,6 +80,15 @@
     return iterable.cursor();
   }
 
+  private static Collection<BasicDBObject> reApplyFilterAfterUnwind(Query query) {
```
suresh-prakash, line 83: This may clearly affect performance.
Can't filter be just done after unwind?

rish691: The filter step before unwind is necessary to reduce the amount of documents unwind is applied to. 
Post filter is not affecting performance but yeah the forced filtering can reduce flexibility of this interface. 
As discussed will introduce a filter inside the UnnestExpression for flexible filtering post each unwind step.

suresh-prakash (approved):

PR #85, opened 2022-06-24
Expose the document path to created time so that it can be used in filters in a generic way

suresh-prakash:
## Description
Expose the document path to created time so that it can be used in filters in a generic way

### Checklist:

### Documentation


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -132,6 +132,11 @@ public boolean healthCheck() {
     return false;
   }
 
+  @Override
+  public String getCreatedTimePath() {
+    return CREATED_AT;
```
kotharironak, line 137: This is the table creation time, is this what are we looking for?

suresh-prakash:
This is not required since Postgres stores the timestamp as Strings.

Issue #82, opened 2022-06-09
Supporting new APIs for postgres implementation


kotharironak:
Recently, there are 4 new APIs were added in document store:

- [x] Aggregate

     - CloseableIterator<Document> aggregate(final org.hypertrace.core.documentstore.query.Query query)

- [x]  Find

    - CloseableIterator<Document> find(final org.hypertrace.core.documentstore.query.Query query)

- [x] count
    - long count(final org.hypertrace.core.documentstore.query.Query query)

- [x] Bulk Operation On Array Value 

   - BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request) throws Exception

- [x] Bulk Update

  - BulkUpdateResult bulkUpdate(List<BulkUpdateRequest> bulkUpdateRequests) throws Exception

For the above APIs, we have implementation for MongoDB, however, we are lacking its support for Postgres. Can we add the support the support?

suddendust:
@kotharironak `count` is already implemented?

kotharironak:
Will take separately bugs + optimization tickets, closing this one.

PR #87, opened 2022-07-04
feat: adds postgres impl for new query interface

kotharironak:

This PR
- handles relation expression in where clause of the new Query interface for `find` + `aggregate` api
- adds a basic unit test for `NEQ` filter
- adds integration test both for `mongo` + `Postgres` for same query

Note:
- all the rest of it is marked as `TODO: not yet implemented` that will follow up as a series of PRs.


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -0,0 +1,205 @@
+package org.hypertrace.core.documentstore.postgres.utils;
```
kotharironak, line 1: Note: Trying to make all these functions common between two independent Query interfaces ( older and newer versions).
So, we will have refactor PRs after the implementation of the new Query Interface.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -0,0 +1,99 @@
...
+    if (having.isPresent()) {
+      sqlBuilder.append(String.format(" %s", pagination.get()));
+    }
+
+    return sqlBuilder.toString();
+  }
+
+  private Optional<String> parseFilter(Optional<FilterTypeExpression> filterTypeExpression) {
+    return Optional.ofNullable(
+        filterTypeExpression.get().accept(new PostgresFilterTypeExpressionVisitor(this)));
```
suresh-prakash, line 68: Missing Optional check for FilterTypeExpression. (`.get()` can throw `NoSuchElementException`)
```suggestion
    return filterTypeExpression.map(
        expression -> expression.accept(new PostgresFilterTypeExpressionVisitor(this)));
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -0,0 +1,205 @@
...
+  private static final String JSON_DATA_ACCESSOR = "->>";
+
+  private static final Set<String> OUTER_COLUMNS =
+      new HashSet<>() {
+        {
+          add(CREATED_AT);
+          add(ID);
+          add(UPDATED_AT);
+        }
+      };
```
suresh-prakash, line 29: Nit:
Could be simplified to 
```suggestion
  private static final Set<String> OUTER_COLUMNS = Set.of(CREATED_AT, ID, UPDATED_AT);
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -1487,6 +1492,34 @@ public void testSearchIteratorInterface(String dataStoreName) throws IOException
...
+                    IdentifierExpression.of("quantity"), NEQ, ConstantExpression.of(10)))
+            .build();
+
+    Iterator<Document> resultDocs = collection.aggregate(query);
+    int count = 0;
+    while (resultDocs.hasNext()) {
+      count++;
+      resultDocs.next();
+    }
+    Assertions.assertEquals(6, count);
```
suresh-prakash, line 1520: Maybe we could have stricter checks (like the exact documents instead of just the count) in a different PR.

suresh-prakash (approved):

suresh-prakash (approved):

PR #90, opened 2022-07-06
feat: add support for selections of new query api

kotharironak:

This is the `third` PR in the series. It takes care of 
- supporting selections (older API always used to return full document)
- adds support for FunctionalType expression
- adds the similar pattern of composing SelectionTypeExpression visitors
- adds different types of Identifier visitor for data accessor, field accessor of JSONB document
- adds basic unit tests for selection
- adds integration test both for `mongo` + `Postgres` for a query involving selection
- handles nested fields scenarios

Note:
There is difference between below two queries:
Q1. `SELECT document->'item' AS item, document->'price' AS price FROM testCollection`
Q2. `SELECT document->'item', document->'price' AS price FROM testCollection`

If, we don't provide an alias name as exp Q1, the name of the return column for jsonb type will be `?column?`. So, as part of the impl, we are defaulting to the field name as an alias if user-defined alias is not provided.

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
@@ -1,7 +1,9 @@
 package org.hypertrace.core.documentstore.postgres;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.util.JSONPObject;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.sql.BatchUpdateException;
@@ -13,6 +15,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Timestamp;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -21,6 +24,7 @@ import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.SneakyThrows;
+import org.apache.commons.lang3.StringUtils;
 import org.hypertrace.core.documentstore.BulkArrayValueUpdateRequest;
 import org.hypertrace.core.documentstore.BulkDeleteResult;
 import org.hypertrace.core.documentstore.BulkUpdateRequest;
@@ -307,7 +311,7 @@ public class PostgresCollection implements Collection {
       ResultSet resultSet = preparedStatement.executeQuery();
       CloseableIterator closeableIterator =
           query.getSelections().size() > 0
-              ? new PostgresResultIteratorWithMetaData(preparedStatement.getMetaData(), resultSet)
+              ? new PostgresResultIteratorWithMetaData(resultSet)
               : new PostgresResultIterator(resultSet);
       return closeableIterator;
     } catch (SQLException e) {
@@ -682,21 +686,32 @@ public class PostgresCollection implements Collection {
   }
 
   static class PostgresResultIteratorWithMetaData extends PostgresResultIterator {
-    protected ResultSetMetaData resultSetMetaData;
 
-    public PostgresResultIteratorWithMetaData(
-        ResultSetMetaData resultSetMetaData, ResultSet resultSet) {
+    public PostgresResultIteratorWithMetaData(ResultSet resultSet) {
       super(resultSet);
-      this.resultSetMetaData = resultSetMetaData;
     }
 
     @Override
     protected Document prepareDocument() throws SQLException, IOException {
+      ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
       int columnCount = resultSetMetaData.getColumnCount();
       Map<String, Object> jsonNode = new HashMap();
       for (int i = 1; i <= columnCount; i++) {
         String columnName = resultSetMetaData.getColumnName(i);
-        jsonNode.put(columnName, MAPPER.readTree(resultSet.getString(i)));
+        String columnValue = resultSet.getString(i);
+        if (StringUtils.isNotEmpty(columnValue)) {
+          columnName = StringUtils.replace(columnName, "_dot_", ".");
+          List<String> keys = Arrays.asList(StringUtils.split(columnName, "."));
+          Collections.reverse(keys);
+          Map<String, Object> pre = new HashMap<>(), cur = pre;
+          pre.put(keys.get(0), MAPPER.readTree(resultSet.getString(i)));
+          for (int j = 1; j < keys.size() - 1; j++) {
+              cur = new HashMap<>();
+              cur.put(keys.get(j), MAPPER.readTree(MAPPER.writeValueAsString(pre)));
+              pre = cur;
+          }
+          jsonNode.put(keys.get(keys.size() - 1), MAPPER.readTree(MAPPER.writeValueAsString(cur)));
+        }
       }
       return new JSONDocument(MAPPER.writeValueAsString(jsonNode));
     }
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
@@ -1,6 +1,5 @@
 package org.hypertrace.core.documentstore.postgres;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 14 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresDataAccessorIdentifierExpressionVisitor.java
```diff
@@ -0,0 +1,43 @@
...
+      PostgresSelectTypeExpressionVisitor baseVisitor, Type type) {
+    super(baseVisitor);
+    this.type = type;
+  }
+
+  @Override
+  public String visit(final IdentifierExpression expression) {
+    String dataAccessor = PostgresUtils.prepareFieldDataAccessorExpr(expression.getName());
+    if (type.equals(Type.NUMERIC)) {
+      return PostgresUtils.prepareCast(dataAccessor, 1);
```
suresh-prakash, line 37: Can we have separate `prepareCast()` methods that will take `Type` as argument instead of simply passing a dummy value of that type (and then inferring back that type)?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -0,0 +1,53 @@
...
+
+    Collector<String, ?, String> collector =
+        getCollectorForFunctionOperator(expression.getOperator());
+
+    String childList =
+        expression.getOperands().stream()
+            .map(exp -> exp.accept(selectTypeExpressionVisitor))
+            .filter(str -> !StringUtils.isEmpty((String) str))
+            .map(str -> "" + str + "")
+            .collect(collector);
```
suresh-prakash, line 35: Nit: Can be simplified to
```suggestion
    String childList =
        expression.getOperands().stream()
            .map(exp -> exp.accept(selectTypeExpressionVisitor))
            .map(Object::toString)
            .filter(StringUtils::isNotEmpty)
            .collect(collector);
```

suresh-prakash: If we also need to perform null checks, then
```suggestion
    String childList =
        expression.getOperands().stream()
            .map(exp -> exp.accept(selectTypeExpressionVisitor))
            .filter(Objects::nonNull)
            .map(Object::toString)
            .filter(StringUtils::isNotEmpty)
            .collect(collector);
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -1,30 +1,80 @@
...
+        new PostgresIdentifierExpressionVisitor();
+
+    String childList =
+        selectionSpecs.stream()
+            .map(
+                selectionSpec ->
+                    Pair.of(
+                        selectionSpec.getExpression().accept(selectTypeExpressionVisitor),
+                        getAlias(selectionSpec, identifierExpressionVisitor)))
+            .filter(p -> !StringUtils.isEmpty((String) p.getLeft()))
```
suresh-prakash, line 62: Just need to make sure if `p.getLeft()` can ever be null. If yes, then, `(String) p.getLeft()` might throw an exception.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -1,30 +1,80 @@
...
+            .map(
+                p ->
+                    StringUtils.isNotEmpty(p.getRight())
+                        ? "" + p.getLeft() + " AS " + p.getRight()
```
suresh-prakash, line 66: Nit: The prefixing empty string could be removed.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -1,30 +1,80 @@
...
+                        : "" + p.getLeft() + "")
```
suresh-prakash, line 67: Nit: Can we prefer `p.getLeft().toString()` instead of adding empty prefix and suffix strings?

Or, is it done intentionally because `p.Left()` could be null.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -1,30 +1,80 @@
...
+
+    // used for if alias is missing
+    PostgresIdentifierExpressionVisitor identifierExpressionVisitor =
+        new PostgresIdentifierExpressionVisitor();
+
+    String childList =
+        selectionSpecs.stream()
+            .map(
+                selectionSpec ->
+                    Pair.of(
```
suresh-prakash, line 59: Can we prefer a local model class over `Pair` to improve readability by avoiding the confusion of what is `p.getLeft()` and what is `p.getRight()`?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -682,24 +684,53 @@ public void close() {
...
+          } else {
+            jsonNode.put(columnName, leafNodeValue);
+          }
+        }
       }
       return new JSONDocument(MAPPER.writeValueAsString(jsonNode));
     }
+
+    private boolean isNestedField(String columnName) {
+      return columnName.contains("_dot_") ? true : false;
```
suresh-prakash, line 713: 1. Any reason we are checking for "_dot_" instead of just "."?
2. Can we define a common static constant for "_dot_"?

suresh-prakash: If we are doing some encoding, can we have a separate Util like `org.hypertrace.core.documentstore.mongo.MongoUtils#encodeKey` for Mongo.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresDataAccessorIdentifierExpressionVisitor.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
kotharironak: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
kotharironak: Done

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: > Any reason we are checking for "dot" instead of just "."?

`.` is not valid char in Alias in SQL. So, replacing it with `_dot_`

> Can we define a common static constant for "dot"?

Done

suresh-prakash (approved):

PR #91, opened 2022-07-08
Fix the bugs in list setters to clear then in case of empty input

suresh-prakash:
## Description
Currently, there is no ability to make the list of selections, sorts, from type expressions and aggregations empty through TransformedQueryBuilder. Fixed this bug.

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/query/Query.java
```diff
@@ -125,7 +125,9 @@ public QueryBuilder setSelection(final Selection selection) {
     }
 
     public QueryBuilder setSelections(final List<SelectionSpec> selectionSpecs) {
-      if (CollectionUtils.isNotEmpty(selectionSpecs)) {
+      if (CollectionUtils.isEmpty(selectionSpecs)) {
```
kotharironak, line 128: Was it crashing? what was the issue?

suresh-prakash: Basically, when we try to set an `emptyList()` as selection list, it was not being set properly.

document-store/src/test/java/org/hypertrace/core/documentstore/query/QueryBuilderTest.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Test
+  void testSetAndClearSelections() {
+    final Query queryWithSelections =
+        queryBuilder.addSelection(ConstantExpression.of("Something")).build();
+    assertEquals(1, queryWithSelections.getSelections().size());
+
+    final Query queryWithoutSelections = queryBuilder.setSelections(emptyList()).build();
+    assertEquals(0, queryWithoutSelections.getSelections().size());
```
kotharironak, line 33: So, previously, this was failing, right?

kotharironak (approved):

document-store/src/test/java/org/hypertrace/core/documentstore/query/QueryBuilderTest.java
suresh-prakash: That's correct.

PR #92, opened 2022-07-08
Fix count aggregation on optional fields

suresh-prakash:
## Description
Fix count aggregation for optional fields in Mongo

### Testing
Added unit and integration tests

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/transformer/MongoSelectionsAddingTransformation.java
```diff
@@ -82,7 +83,7 @@ public Optional<SelectionSpec> visit(final AggregateExpression expression) {
     final String encodedAlias = encodeKey(alias);
     final SelectTypeExpression pairingExpression;
 
-    if (expression.getAggregator() == DISTINCT_COUNT) {
+    if (expression.getAggregator() == DISTINCT_COUNT || expression.getAggregator() == COUNT) {
```
skjindal93, line 86: Can we add a comment that, the hack can be removed once we upgrade to MongoDB 5.0, with the appropriate MongoDB issue link (if any)?

suresh-prakash: The closest, I found is https://stackoverflow.com/questions/24182379/mongodb-multiple-count-with-exists, but this doesn't count fields which exist but, have a `0` or `false` value.

skjindal93 (approved):

PR #95, opened 2022-07-11
feat: add supports for aggregation filter expression for new query api

kotharironak:

This is the `fifth` PR in the series. It takes care of 
- supporting aggregation filter
- refactor a code to take PostgresQueryParser as the running object
- adds support for memorizing alias expressions in PostgresQueryParser
- adds unit test for having clause
- adds integration test for having clause

Note:
We can't use the alias in WHERE / HAVING clause in SQL expression. So, to support it, we need to rewrite the alias expression.

As an example, below is not a valid query:
```
SELECT COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) AS qty_count, document->'item' AS item 
FROM testCollection 
GROUP BY document->'item' 
HAVING qty_count <= 10
```

However, we can re-write it as:
```
SELECT COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) AS qty_count, document->'item' AS item 
FROM testCollection 
GROUP BY document->'item' 
HAVING COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) <= 10
```

### DIFF
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
@@ -1531,7 +1531,7 @@ public class DocStoreTest {
             .build();
 
     Iterator<Document> iterator = collection.aggregate(query);
-    assertSizeAndDocsEqual(dataStoreName, iterator, 6, "mongo/simple_filter_quantity_neq_10.json");
+    assertSizeAndDocsEqual(dataStoreName, iterator, 4, "mongo/simple_filter_quantity_neq_10.json");
   }
 
   @ParameterizedTest
@@ -1685,6 +1685,35 @@ public class DocStoreTest {
     Iterator<Document> resultDocs = collection.aggregate(query);
     assertSizeAndDocsEqual(
         dataStoreName, resultDocs, 3, "mongo/test_aggregation_expression_result.json");
+  }
+
+
+  @ParameterizedTest
+  @MethodSource("databaseContextProvider")
+  public void testQueryV1AggregationFilter(String dataStoreName) throws IOException {
+    Map<Key, Document> documents = createDocumentsFromResource("mongo/collection_data.json");
+    Datastore datastore = datastoreMap.get(dataStoreName);
+    Collection collection = datastore.getCollection(COLLECTION_NAME);
+
+    // add docs
+    boolean result = collection.bulkUpsert(documents);
+    Assertions.assertTrue(result);
+
+    org.hypertrace.core.documentstore.query.Query query =
+        org.hypertrace.core.documentstore.query.Query.builder()
+            .addSelection(
+                AggregateExpression.of(DISTINCT_COUNT, IdentifierExpression.of("quantity")),
+                "qty_count")
+            .addSelection(IdentifierExpression.of("item"))
+            .addAggregation(IdentifierExpression.of("item"))
+            .setAggregationFilter(
+                RelationalExpression.of(
+                    IdentifierExpression.of("qty_count"), LTE, ConstantExpression.of(10)))
+            .build();
+
+    Iterator<Document> resultDocs = collection.aggregate(query);
+    assertSizeAndDocsEqual(
+        dataStoreName, resultDocs, 4, "mongo/distinct_count_response.json");
   }
 
   @ParameterizedTest
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
@@ -1,8 +1,10 @@
 package org.hypertrace.core.documentstore.postgres.query.v1;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
-import org.hypertrace.core.documentstore.expression.type.FilterTypeExpression;
+import lombok.Getter;
 import org.hypertrace.core.documentstore.postgres.Params;
 import org.hypertrace.core.documentstore.postgres.Params.Builder;
 import org.hypertrace.core.documentstore.postgres.query.v1.vistors.PostgresFilterTypeExpressionVisitor;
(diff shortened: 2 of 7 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -1,104 +1,112 @@
...
 
-    // selection clause
-    Optional<String> selectionClause = parseSelection(query.getSelections());
-    if (selectionClause.isPresent()) {
-      sqlBuilder = new StringBuilder();
-      sqlBuilder.append(String.format("SELECT %s FROM %s", selectionClause.get(), collection));
-    }
     // where clause
-    Optional<String> whereFilter = parseFilter(query.getFilter());
+    Optional<String> whereFilter = parseFilter();
```
kotharironak, line 44: Note: we need alias only in aggregation filter, so where clause is parsed first, and alias is prepared later in selection clause for aggregation filter.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -215,6 +215,76 @@ public static String parseNonCompositeFilter(
     return filters;
   }
 
+  public static String prepareParsedNonCompositeFilter(
```
kotharironak, line 218: Note: we can combine this method with the existing `parseNonCompositeFilter`. However, I had to add a few if-clauses. So, I will do the clean (or refactor) as follow-up.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -114,7 +114,7 @@ private static String prepareParameterizedStringForList(
 
   public static String parseNonCompositeFilter(
       String fieldName, String op, Object value, Builder paramsBuilder) {
-    String fullFieldName = prepareCast(prepareFieldDataAccessorExpr(fieldName), value);
+    String fullFieldName = prepareCast(prepareFieldAccessorExpr(fieldName).toString(), value);
```
kotharironak, line 117: In SQL, both having clause and group by clause should match the field accessor pattern. This has been changed to support normal field expression other than functional/aggregation in having cluase.

See the below query. Chec how `price` is accessed using `->` instead of `->>`.

```
SELECT COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) AS qty_count, 
document->'item' AS item, 
document->'price' AS price 
FROM testCollection 
GROUP BY document->'item',document->'price' 
HAVING (COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) <= 10) AND (
CAST (document->'price' AS NUMERIC) > 5)
```

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
```diff
@@ -41,7 +42,7 @@ void testParseSimpleFilter() {
     String sql = postgresQueryParser.parse(query);
     Assertions.assertEquals(
         "SELECT * FROM testCollection "
-            + "WHERE document->'quantity' IS NULL OR CAST (document->>'quantity' AS NUMERIC) != ?",
+            + "WHERE document->'quantity' IS NULL OR CAST (document->'quantity' AS NUMERIC) != ?",
```
kotharironak, line 45: See the comments above for all the test case related changes in this change - https://github.com/hypertrace/document-store/pull/95/files#r917602631

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -95,7 +95,7 @@ public static String prepareCast(String field, Type type) {
     } else if (type.equals(Type.BOOLEAN)) {
       return String.format(fmt, field, type);
     } else /* default is string */ {
-      return field;
+      return String.format(fmt, field, type);
```
kotharironak, line 98: Now, w/o data accessor pattern, we need a cast for string fields too. Pl. see the test case `testFilterWithNestedFiled`

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -1687,6 +1710,71 @@ public void testQueryV1AggregationExpression(String dataStoreName) throws IOExce
...
+  @ParameterizedTest
+  @MethodSource("databaseContextProvider")
+  public void testQueryV1AggregationFilter(String dataStoreName) throws IOException {
+    Map<Key, Document> documents = createDocumentsFromResource("mongo/collection_data.json");
+    Datastore datastore = datastoreMap.get(dataStoreName);
+    Collection collection = datastore.getCollection(COLLECTION_NAME);
+
+    // add docs
+    boolean result = collection.bulkUpsert(documents);
+    Assertions.assertTrue(result);
```
suresh-prakash, line 1722: Maybe in a separate PR, could we make a method for lines from 1716 to 1722 as adding documents part seems to be repeated.

suresh-prakash: Also, if possible, as an optimisation, we could move all the query tests into a single class with a `setUp()` (@BeforeAll) method that inserts documents just once.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -1,104 +1,114 @@
...
-    this.collection = collection;
+  @Getter private Builder paramsBuilder;
+
+  @Getter private Query query;
+
+  // map of alias name to parsed expression
+  // e.g qty : COUNT(DISTINCT CAST(document->>'quantity' AS NUMERIC))
+  @Getter private Map<String, String> pgSelections;
+
+  private void resetParser(Query query) {
```
suresh-prakash, line 29: Nit: 
This looks like a setter method for `query` with some custom logic.
Can this method be named `setQuery` instead?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -1,104 +1,114 @@
...
+    this.query = query;
+    paramsBuilder = Params.newBuilder();
```
suresh-prakash, line 31: This makes this class thread "unsafe".
If a thread calls `resetParser` while the other thread is using `paramsBuilder` or `pgSelections`, it would end up in undesired concurrency issues.

Instead can we make sure to initialise `paramsBuilder`, `query` and `pgSelections` only once (in the constructor) and make these fields `final`?

That way, no 2 different threads can mutate the same `PostgresQueryParser` object (and the client code is forced to create a different instance for each parsing).

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -1,104 +1,114 @@
...
-  private Optional<String> parseSelection(List<SelectionSpec> selectionSpecs) {
-    return Optional.ofNullable(PostgresSelectTypeExpressionVisitor.getSelections(selectionSpecs));
+  private Optional<String> parseSelection() {
+    return Optional.ofNullable(PostgresSelectTypeExpressionVisitor.getSelections(this));
   }
 
-  private Optional<String> parseFilter(Optional<FilterTypeExpression> filterTypeExpression) {
-    return filterTypeExpression.map(
-        expression -> expression.accept(new PostgresFilterTypeExpressionVisitor(this)));
+  private Optional<String> parseFilter() {
```
suresh-prakash, line 90: This (avoiding Optional method parameters) is a clean improvement.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -40,14 +42,38 @@ public String visit(final RelationalExpression expression) {
...
-    return PostgresUtils.parseNonCompositeFilter(
-        key, operator.toString(), value, this.postgresQueryParser.getParamsBuilder());
+    return postgresQueryParser.getPgSelections().containsKey(key)
+        ? PostgresUtils.prepareParsedNonCompositeFilter(
+            postgresQueryParser.getPgSelections().get(key),
+            operator.toString(),
+            value,
+            this.postgresQueryParser.getParamsBuilder())
+        : PostgresUtils.parseNonCompositeFilter(
+            key, operator.toString(), value, this.postgresQueryParser.getParamsBuilder());
```
suresh-prakash, line 58: This logic makes it hard to understand without explicit comments.
One way is to add comments explaining why we are doing so.
Another way is to transform the query before parsing (just like we do it for Mongo @ org.hypertrace.core.documentstore.mongo.query.transformer.MongoQueryTransformer).

I prefer the 2nd approach because of 

1. Better debugging (ability to log the original and the transformed query)
2. Can be extended to support all the unsupported querying patterns. [One example is https://github.com/hypertrace/document-store/pull/92, where we support an unsupported operation by transforming the query into a supported form with minimal changes to the parsing logic].
3. Single-class single responsibility. The visitor classes will just be responsible for parsing (translating) the query as it is given, while the transformer classes will be doing all the query transformations.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -45,13 +47,15 @@ public <T> T visit(final IdentifierExpression expression) {
     return baseVisitor.visit(expression);
   }
 
-  @AllArgsConstructor
+  @NoArgsConstructor
```
suresh-prakash, line 50: Should we mark this is `@Value` class or at least have the fields immutable?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
```diff
@@ -60,13 +64,25 @@ public static String getSelections(final List<SelectionSpec> selectionSpecs) {
...
+                      selectionSpec.getExpression().accept(selectTypeExpressionVisitor);
+                  if (!StringUtils.isEmpty(selectionSpec.getAlias())) {
+                    pgSelection.alias = selectionSpec.getAlias();
+                    pgSelections.put(pgSelection.alias, pgSelection.fieldName);
+                  } else {
+                    pgSelection.alias =
+                        PostgresUtils.encodeAliasForNestedField(
+                            selectionSpec.getExpression().accept(identifierExpressionVisitor));
+                  }
+                  return pgSelection;
```
suresh-prakash, line 84: Too many things happening inside a lambda expression used inside `map`. Can we break it into a methods of its own?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
suresh-prakash: Can we also add a code comment mentioning the need for casting strings?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
suresh-prakash: Probably, adding this explanation in a comment (maybe here or at the top of the class, or in tests) would help.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
suresh-prakash: Can we add a `// TODO:` comment so we can keep track?

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
```diff
@@ -299,4 +330,73 @@ void testAggregationExpressionDistinctCount() {
...
+    String sql = postgresQueryParser.parse(query);
+
+    Assertions.assertEquals(
+        "SELECT COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) AS qty_count, "
+            + "document->'item' AS item, "
+            + "document->'price' AS price "
+            + "FROM testCollection "
+            + "GROUP BY document->'item',document->'price' "
+            + "HAVING (COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) <= ?) "
+            + "AND (CAST (document->'price' AS NUMERIC) > ?)",
```
suresh-prakash, line 395: Ideally, this `(CAST (document->'price' AS NUMERIC) > ?)` can be a `WHERE` clause filter.

Can we test with both `WHERE` and `HAVING` filters. If we want to testing `HAVING` with composition, can perform `AND` with `(COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) > ?)` so it becomes similar to a `BETWEEN` filter.

suresh-prakash (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
kotharironak: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
kotharironak: Done.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: Ya, we will have to clean up tests. Planning to do as follow-up.

suresh-prakash: Sure. Makes sense.
In general, for any potential follow-ups, could we just add a single `TODO` comment so that we can keep track?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
kotharironak: As this parser needs to be used behind the collection interfaces, it should be fine. Yah, was planning to do that. 
Done, as part of this PR.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
kotharironak: No need due to - https://github.com/hypertrace/document-store/pull/95#discussion_r917748266

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
kotharironak: I will do this as the follow-up, first want to identify all the required things for functionality. Added todo.

suresh-prakash: Sure. 👍

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
kotharironak: Yeah, but our query expression doesn't restrict it, and the client can prepare this way too.

suresh-prakash: That's right. We don't restrict. Just wanted to have a test case with both `HAVING` and `WHERE` filters as well.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresSelectTypeExpressionVisitor.java
kotharironak: Done. Refactored a bit.

suresh-prakash (approved):

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
kotharironak: Done, added the test.

PR #96, opened 2022-07-11
feat: add support for orderby clause for new query api

kotharironak:

This is the `sixth` PR in the series. It takes care of 
- supporting order by clause
- as alias is supported in order by clause in SQL, it uses it if an alias exists
- for non-alias expression, prepare the identifier expression using the field accessor pattern
- adds unit test + integration tests

Note:
Alias is supported in the ORDER BY clause of SQL.

An example query using alias field as well as a non-alias field:
```
SELECT COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) AS qty_count, 
document->'item' AS item 
FROM testCollection 
GROUP BY document->'item' 
HAVING COUNT(DISTINCT CAST (document->>'quantity' AS NUMERIC) ) <= 10 
ORDER BY qty_count DESC,document->'item' DESC
```


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutorIntegrationTest.java
```diff
@@ -498,27 +497,6 @@ public void testAggregateWithMultipleGroupingLevels() throws IOException {
...
-            .setAggregationFilter(
-                RelationalExpression.of(
-                    IdentifierExpression.of("qty_count"), LTE, ConstantExpression.of(1000)))
-            .addSort(IdentifierExpression.of("qty_count"), DESC)
-            .addSort(IdentifierExpression.of("item"), DESC)
-            .build();
-
-    Iterator<Document> resultDocs = collection.aggregate(query);
-    assertDocsEqual(resultDocs, "mongo/distinct_count_response.json");
-    assertSizeEqual(query, "mongo/distinct_count_response.json");
```
suresh-prakash, line 519: This call is testing the `count()` API, which is missing in the new implementation.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/Utils.java
```diff
@@ -92,4 +96,20 @@ public static Map<String, Object> convertDocumentToMap(Document document)
...
+
+    int actualSize = 0;
+    List<Map<String, Object>> actual = new ArrayList<>();
+    while (documents.hasNext()) {
+      actual.add(convertDocumentToMap(documents.next()));
+      actualSize++;
+    }
+
+    assertEquals(expected, actual);
+    assertEquals(expectedSize, actualSize);
```
suresh-prakash, line 113: For every `aggregate()` result, can we also validate that the size matches the `count()` API? Ref.: `org.hypertrace.core.documentstore.mongo.MongoQueryExecutorIntegrationTest#assertSizeEqual(Query, java.lang.String)`

suresh-prakash (approved):

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/Utils.java
kotharironak: I am counting the number of documents as part of the API - assertDocsAndSizeEqual

suresh-prakash: Right! But, it is not hitting the `Collection.count()` method.
Therefore, the functionality of `count()` is not tested along with `aggregate()`.

kotharironak: I have kept a Collection.count as separate API impl - https://github.com/hypertrace/document-store/issues/82
While doing that, we will add the tests.

PR #97, opened 2022-07-11
Postgres Bulk Update Array

suddendust:
- SET
- ADD
- REMOVE

All extended `Query` to allow chaining.


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
-  public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request) {
-    throw new UnsupportedOperationException();
+  public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request)
+      throws Exception {
+    Operation operation = request.getOperation();
+    switch (operation) {
+        // todo: All of these implementations do a GET -> in memory update -> UPSERT which is not
+        // optimal. Changes this logic to use a single query using jsob_set and json_insert
+      case ADD:
+        return addImpl(request);
```
kotharironak, line 279: nit:For readability, we can rename the function names>

`addImpl` -> `bulkAddOnArrayValue`
`setImpl` -> `bulkSetOnArrayValue`
`removeImpl` -> bulkRemoveOnArrayValue`

above suggestion is based on the pattern - bulk<Operation>OnArrayValue

What do you think?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        throw new UnsupportedOperationException("Unsupported operation: " + operation);
+    }
+  }
+
+  private BulkUpdateResult removeImpl(BulkArrayValueUpdateRequest request) throws IOException {
+    Set<Key> keys = request.getKeys();
+    String[] pathTokens = request.getSubDocPath().split("\\.");
+    Map<Key, Document> upsertMap = new HashMap<>();
+    for (Key key : keys) {
+      Query getQuery = getSelectDocumentQueryForKey(key);
```
kotharironak, line 294: We can use a single Query to fetch all documents using the `IN` filter?

```
Select document from <> where ID in (<v1>, <v2>,... )
```

Any specific reason for doing iteratively?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        } catch (JsonProcessingException e) {
+          LOGGER.error("Malformed JSON for key: {}", key);
+          continue;
+        }
+        JsonNode currNode = rootNode;
+        for (String pathToken : pathTokens) {
+          currNode = currNode.path(pathToken);
+        }
+        // if we do not end up in an array node, then the path doesn't exist. So do not do anything
+        if (!currNode.isArray()) {
```
kotharironak, line 310: Can `currNode` be null if the path doesn't exist?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+  private BulkUpdateResult removeImpl(BulkArrayValueUpdateRequest request) throws IOException {
+    Set<Key> keys = request.getKeys();
+    String[] pathTokens = request.getSubDocPath().split("\\.");
+    Map<Key, Document> upsertMap = new HashMap<>();
+    for (Key key : keys) {
+      Query getQuery = getSelectDocumentQueryForKey(key);
+      CloseableIterator<Document> searchRes = search(getQuery);
+      if (searchRes.hasNext()) {
+        JsonNode rootNode;
+        try {
```
kotharironak, line 298: we can use the method - `getSubDocAsJSON(<key from request>,  doc)`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        }
+        // create path if missing
+        JsonNode arrayNode = createPathInDoc(request.getSubDocPath(), rootNode);
+        ArrayNode candidateArray = (ArrayNode) arrayNode;
+        for (Document subDoc : request.getSubDocuments()) {
+          // we don't insert docs if they exist already (so the array behaves like a set here)
+          boolean alreadyContainsDoc = false;
+          Iterator<JsonNode> arrayElems = candidateArray.elements();
+          JsonNode subDocAsJSON = getSubDocAsJSON(key, subDoc);
+          while (arrayElems.hasNext()) {
```
kotharironak, line 361: Can we do this using `set` operations?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
suddendust: `.path` never returns null, it returns a `MissingNode` instead. So this condition will evaluate to be false if path doesn't exist.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
suddendust: Missed this, will fix.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        }
+        upsertMap.put(key, new JSONDocument(rootNode));
+      } else {
+        LOGGER.warn("Could not get row with key: {}", key);
+      }
+    }
+    return upsertDocs(upsertMap);
+  }
+
+  private JsonNode getSubDocAsJSON(Key key, Document subDoc) throws IOException {
```
kotharironak, line 380: Instead of throwing, would returning null will suffice? So, that way places where you want to continue, will be avoided.

```
private JsonNode convertDocToJson(Document subDoc) {
    JsonNode subDocAsJSON;
    try {
      subDocAsJSON = MAPPER.readTree(subDoc.toJson());
    } catch (Exception e) {
      LOGGER.error("Malformed JSON", e);
      return null;
    }
    return subDocAsJSON;
  }
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +269,166 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
...
+        JsonNode currNode = rootNode;
+        for (String pathToken : pathTokens) {
+          currNode = currNode.path(pathToken);
+        }
+        // if we do not end up in an array node, then the path doesn't exist. So do not do anything
+        if (!currNode.isArray()) {
+          continue;
+        }
+        ArrayNode candidateArray = (ArrayNode) currNode;
+        for (Document subDoc : request.getSubDocuments()) {
```
kotharironak, line 314: Would using `set` operation work? Something like below?

```
Set<JsonNode> itemsToRemove = request.getSubDocuments()
            .stream().map(subDoc -> convertDocToJson(subDoc))
        .collect(Collectors.toUnmodifiableSet());

        Set<JsonNode> existingItems = new HashSet<>();
        Iterator<JsonNode> iterator = candidateArray.iterator();
        while (iterator.hasNext()) {
          existingItems.add(iterator.next());
        }

        existingItems.removeAll(itemsToRemove);
        candidateArray.removeAll();
        candidateArray.addAll(existingItems);

        upsertMap.put(key, new JSONDocument(rootNode));
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
suddendust: @kotharironak Actually did this to keep the behaviour similar to Mongo. Mongo currently throws a `JSONParseException` if the docs are malformed.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -263,8 +268,21 @@ public BulkUpdateResult bulkUpdateSubDocs(Map<Key, Map<String, Document>> docume
   }
 
   @Override
-  public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request) {
-    throw new UnsupportedOperationException();
+  public BulkUpdateResult bulkOperationOnArrayValue(BulkArrayValueUpdateRequest request)
+      throws Exception {
+    Operation operation = request.getOperation();
+    switch (operation) {
+        // todo: All of these implementations do a GET -> in memory update -> UPSERT which is not
```
kotharironak, line 275: Let's add this as a method-level comment, explaining the approach.

/**The approach was taken for in-memory join as there is no direct support for set operation on JSON array. If, needed, this can be optimized or implemented using jsonb_set or jsonb_insert*/

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -564,6 +512,250 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
     return false;
   }
 
+  @Override
+  public CloseableIterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Document> documents)
```
kotharironak, line 516: Where are we using this one?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -564,6 +512,250 @@ public boolean bulkUpsert(Map<Key, Document> documents) {
...
+
+  private Map<String, String> getDocIdToTenantIdMap(BulkArrayValueUpdateRequest request) {
+    return request.getKeys().stream()
+        .collect(
+            Collectors.toMap(
+                key -> key.toString().split(":")[1], key -> key.toString().split(":")[0]));
+  }
+
+  private BulkUpdateResult bulkAddOnArrayValue(BulkArrayValueUpdateRequest request)
+      throws IOException {
```
kotharironak, line 633: As discussed:
- can we move common code to the base method?
- can we use the same pattern of reading `existingElems` in all the three methods - prefer `foreach` vs `while`?
- remove `asserts` and assertions issue of docstore test (different result between mongo vs postgres).

Can you file a new issue for `PostgresResultIteratorWithMetaData` for handling `string` columns?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -873,6 +880,853 @@ public void testIgnoreCaseLikeQuery(String dataStoreName) throws IOException {
...
+        Utils.createDocument(ImmutablePair.of("value", Map.of("string", "Label3")));
+    // malformed doc
+    Document label3Document = () -> "malformedJson";
+    List<Document> subDocuments = List.of(label2Document, label3Document);
+
+    BulkArrayValueUpdateRequest bulkArrayValueUpdateRequest =
+        new BulkArrayValueUpdateRequest(
+            Set.of(key1), "attributes.labels.valueList.values", SET, subDocuments);
+    // candidate under test
+    try {
```
kotharironak, line 1117: we can use `assertThrows` here.

suddendust: Done

kotharironak (approved):
Overall looks good to me.
Can you also take care of `test_bulkOperationOnArrayValue_addOperation` test of `MogoDocStoreTest.Java` in follow up PR. I guess, you just have to delete it.

kotharironak:
This PR handles bulk update on array values of ticket - https://github.com/hypertrace/document-store/issues/82

PR #100, opened 2022-07-13
feat: add support for unnest expression

kotharironak:

This is the `eighth` PR in the series. It takes care of 
- supporting unnest expression
- data access parser and field access parser takes input for column
- passes Postgres parser object to selection type expression visitor
- handles both preserve empty and null array for true and false
- handles unnest expression filter clause
- handles main filter clause
- adds unit test + integration tests

**Approch:**
In postgres, when we unwind array fields, it adds new jsonb columns. So if there is a query that involves reading values from unwind column field, it needs right mapping in the rest of the SQL expression.

As an example of the following document:
```
{
    "name" : "mr.a",
    "adress: [{"type":"home", "street":"h1"}, {"type":"office", "street":"o1"}]
}
```
Unwind on `address` field will make the below table:

| document | address |
| --- | ----------- |
| { "name" : "mr.a", "adress: [{"type":"home", "street":"h1"}, {"type":"office", "street":"o1"}]} | {"type":"home", "street":"h1"} |
| { "name" : "mr.a", "adress: [{"type":"home", "street":"h1"}, {"type":"office", "street":"o1"}]} | {"type":"office", "street":"o1"} |

So, for the below query, which involves selecting data from `address` field, we need to use the `address` jsonb column not the `document` column
```
org.hypertrace.core.documentstore.query.Query query =
        org.hypertrace.core.documentstore.query.Query.builder()
            .addSelection(IdentifierExpression.of("name"))
            .addSelection(IdentifierExpression.of("address.type"))
            .addFromClause(UnnestExpression.of(IdentifierExpression.of("address"), false))
            .build();
```

This was taken care of by keeping the map of column names against the unnest expression field. And, using that information in query parsing.

Translated query (assuming base table name as `mainTable` that contains `document` jsonb):
```
With 
table1 as (SELECT * from mainTable), 
table2 as (SELECT * FROM table1, jsonb_array_elements(document->'address') p1(address)) 
SELECT document->'name' AS item, address->'type' AS address_dot_type FROM table2
```

**Note**
- alias with the capital case is not handled as part of this PR
- if `preserveNullAndEmptyArrays` has true, there are a few discrepancies with `empty` array data between mongo and Postgres the way it handles. So, a few test cases are not yet moved. We will see if we can manage it.


document-store/src/integrationTest/resources/mongo/aggregate_on_nested_array_reponse.json
```diff
@@ -5,22 +5,22 @@
         "type": "online"
       }
     },
-    "totalSales": 9000
+    "totalsales": 9000
```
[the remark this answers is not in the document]
suresh-prakash: Hope we've marked a TODO somewhere

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FieldToPgColumnTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+
+  public FieldToPgColumnTransformer(PostgresQueryParser postgresQueryParser) {
+    this.postgresQueryParser = postgresQueryParser;
+  }
+
+  public FieldToPgColumn transform(String orgFieldName) {
+    Optional<String> parentField =
+        postgresQueryParser.getPgColumnNames().keySet().stream()
+            .filter(f -> orgFieldName.startsWith(f))
+            .max(
```
suresh-prakash, line 21: Can be simplified to `.max(Comparator.comparingInt(String::length))`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FieldToPgColumnTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+  private PostgresQueryParser postgresQueryParser;
+
+  public FieldToPgColumnTransformer(PostgresQueryParser postgresQueryParser) {
+    this.postgresQueryParser = postgresQueryParser;
+  }
+
+  public FieldToPgColumn transform(String orgFieldName) {
+    Optional<String> parentField =
+        postgresQueryParser.getPgColumnNames().keySet().stream()
+            .filter(f -> orgFieldName.startsWith(f))
```
suresh-prakash, line 20: 1. Are we saying that the alias can only be a part of the field name?
2. Is it the reason for handling only small cases?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -22,18 +23,20 @@ public class PostgresUtils {
   private static final String DOT_STR = "_dot_";
   private static final String DOT = ".";
 
-  private static final Set<String> OUTER_COLUMNS = Set.of(CREATED_AT, ID, UPDATED_AT);
+  public static final Set<String> OUTER_COLUMNS =
+      new TreeSet<>(List.of(ID, CREATED_AT, UPDATED_AT));
```
suresh-prakash, line 27: Any reason for preferring a TreeSet over the normal set?

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
```diff
@@ -542,4 +543,112 @@ void testFindWithSortingAndPagination() {
...
+            .addFromClause(UnnestExpression.of(IdentifierExpression.of("sales"), true))
+            .addFromClause(UnnestExpression.of(IdentifierExpression.of("sales.medium"), true))
+            .build();
+
+    PostgresQueryParser postgresQueryParser = new PostgresQueryParser(TEST_COLLECTION, query);
+    String sql = postgresQueryParser.parse();
+
+    Assertions.assertEquals(
+        "With \n"
+            + "table1 as (SELECT * from testCollection), \n"
```
suresh-prakash, line 595: table1 seems unused. Can we get rid of this temporary table?

document-store/src/integrationTest/resources/mongo/aggregate_on_nested_array_reponse.json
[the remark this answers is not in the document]
kotharironak: No need for ToDo, as part of closing the ticket, it's required, and currently working on it as the next PR.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FieldToPgColumnTransformer.java
kotharironak: No, even if I pass the aliases to SQL query, the return column name from Postgresql is in lowerCase as default.
e.g
`Select document->'item' as ITEM from myTest` will return me `item` as a column. So, the Document iterator has no knowledge of lower/upper case while parsing the result. Prior (Query API v0), we were not supporting selection, so there was no need. But, with the new Query interface, we need to support alias with either case. As this is independent of `UnnestExpression`,  looking at it as a separate PR. Working on it.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
kotharironak: No specific reason, wanted to have generated fields in query same pattern. But, that's not the mandatory requriement.

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
kotharironak: `table2 as (SELECT * FROM table1, jsonb_array_elements(document->'sales') p1(sales)`

Need it when there is a `WHERE` caluse as it should apply before unwind.

suresh-prakash (approved):

PR #103, opened 2022-07-18
mongo tests updated

anubhav-sharma13:
## Description
In our previous iteration, we refactored multiple test cases for both mongo and postgres. So this PR is for removing and checking the tests from individual files.


anubhav-sharma13:
@kotharironak  please review

PR #106, opened 2022-07-19
feat: add support for count api for new query

kotharironak:

This is the `tenth` PR in the series. It takes care of 
- implements count API 
- moves all the test cases to common
- marked un-supported tests under `databaseContextMongo`, might need client change.


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -2357,582 +2313,6 @@ public void testSearchIteratorInterface(String dataStoreName) throws IOException
     }
   }
 
-  @ParameterizedTest
-  @MethodSource("databaseContextProvider")
```
kotharironak, line 2361: Moved all-new query v1 related API test cases to `DocStoreQueryV1Test.java`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -345,7 +345,16 @@ public CloseableIterator<Document> aggregate(
 
   @Override
   public long count(org.hypertrace.core.documentstore.query.Query query) {
-    throw new UnsupportedOperationException();
+    long count = 0;
+    try (CloseableIterator<Document> iterator = aggregate(query)) {
```
suresh-prakash, line 349: This implementation is not efficient as it fetches the entire result-set from the database host to the client host. Ideally, we would need something like `SELECT COUNT(*) FROM <aggregation_sub_query>`.
Can be handled later, though.

suresh-prakash (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: we need to see if we can directly get count only via postgres that would be better.

In the current form, as the API takes `Query` as input, it can accept any complex query.
So, a query like the below will error out if we replace selection clause with `count (*)`, but it can work in mongo impl:

```
SELECT 
count(*)
FROM myTest WHERE document->'quantity' IS NULL OR CAST (document->>'quantity' AS NUMERIC) != 10 
GROUP BY document->'item' 
HAVING (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) >= 11) 
AND (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) <= 99) 
ORDER BY "total" DESC OFFSET 0 LIMIT 10
```

And for valid queries, we still need the iterator with the results for the `count(*)` approach:
```
SELECT 
count(*)
FROM myTest WHERE document->'quantity' IS NULL OR CAST (document->>'quantity' AS NUMERIC) != 10 
GROUP BY document->'item' 
HAVING (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) >= 11) 
AND (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) <= 99) 
OFFSET 0 LIMIT 10
```
The output of the above query in Postgres will be multiple rows as it involves group_by:
| count |
|------|
| 1 |
| 1 |

However, in the above case, we can avoid the computation of other selection fields. On a different opinion, I hardly see the need for this API where we have a generic aggregate API. The client can form the valid `count(*)` query using generic `Query` if it's looking for a count.

suresh-prakash: There's a difference between
```
SELECT 
count(*)
FROM myTest WHERE document->'quantity' IS NULL OR CAST (document->>'quantity' AS NUMERIC) != 10 
GROUP BY document->'item' 
HAVING (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) >= 11) 
AND (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) <= 99) 
OFFSET 0 LIMIT 10
```
and 
```
SELECT COUNT(*) FROM 
(SELECT 
count(*)
FROM myTest WHERE document->'quantity' IS NULL OR CAST (document->>'quantity' AS NUMERIC) != 10 
GROUP BY document->'item' 
HAVING (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) >= 11) 
AND (SUM( CAST (document->>'price' AS NUMERIC) * CAST (document->>'quantity' AS NUMERIC) ) <= 99) 
OFFSET 0 LIMIT 10) AS tempTable
```

The count API is not an aggregation specific counting. It is the count of the number of aggregations themselves.

> we still need the iterator with the results for the count(*) approach

Not required if we count the number of aggregations.

> do we need this API where we have a generic aggregate API? The client can form the valid count query using generic Query if it's looking for a count. So, instead of having three different APIs - find, aggregate, and count, we can have a single API that takes Query as input.

Agree that `find` and `aggregate` could be combined. I still see a need for `count` because, it is the count of aggregations and not the count as a result of aggregation.

kotharironak: Yes, the count requirement on DocumentSet makes sense, however, I think it should be independent whether it's a standard set or an aggregated set.

Secondly, if we see it differently, it is one of the requirements on derived/aggregated/transformed DocumentSet from the base set. We might need distinct_count on top of it. (e.g we have added different APIs for specific needs in the update). So, here, we are looking for chaining requirements. Now, we also have support for FromClause in our Query language, I was thinking that we can extend it to support in `Query` as part of it (`select count(*) from (Query)`). That way, we can make it generic. 

As we already added this API, I have wrapped it using the sub-query as part of this PR - https://github.com/hypertrace/document-store/pull/107/. But, let's see if there are more requirements, and if we can take the above approach of extending from the clause as part of the `find` API.

PR #107, opened 2022-07-20
fix: fixed few more mongo specific tests

kotharironak:

This is the follow-up PR. It takes care of,
- fixed all the tests that were marked only for mongo context except one.
    -  DISTINCT with Groupby is not supported on postgres impl as of now.
- allows fields accessor along with expression in having clause (aggregation filter)
- rewrite count API using sub-query as discussed here - https://github.com/hypertrace/document-store/pull/106#discussion_r926563000

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregateExpressionVisitor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregateExpressionVisitor.java
@@ -52,7 +52,7 @@ public class PostgresAggregateExpressionVisitor extends PostgresSelectTypeExpres
             ? stringTypeDataAccessorVisitor
             : numericTypeDataAccessorVisitor;
 
-    String value = expression.getExpression().accept(selectTypeExpressionVisitor);
+    String value = expression.getExpression().accept(selectTypeExpressionVisitor).toString();
     return value != null ? convertToAggregationFunction(operator, value) : null;
   }
 
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregationFilterTypeExpressionVisitor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregationFilterTypeExpressionVisitor.java
@@ -11,8 +11,15 @@ import org.hypertrace.core.documentstore.postgres.utils.PostgresUtils;
 public class PostgresAggregationFilterTypeExpressionVisitor
     extends PostgresFilterTypeExpressionVisitor {
 
+  private PostgresSelectTypeExpressionVisitor lhsFieldVisitor;
+  private PostgresSelectTypeExpressionVisitor lhsVisitor;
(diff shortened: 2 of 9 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregateExpressionVisitor.java
```diff
@@ -52,7 +52,7 @@ public String visit(final AggregateExpression expression) {
             ? stringTypeDataAccessorVisitor
             : numericTypeDataAccessorVisitor;
 
-    String value = expression.getExpression().accept(selectTypeExpressionVisitor);
+    String value = expression.getExpression().accept(selectTypeExpressionVisitor).toString();
```
kotharironak, line 55: Fixed the issue for`ContantExpression` having a numeric value.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresAggregationFilterTypeExpressionVisitor.java
```diff
@@ -22,27 +29,22 @@ public String visit(final RelationalExpression expression) {
...
-    // In SQL, the alias is not supported in the Filter and Having clause.
-    // As of now, where is clause is first parsed and it is not using any alias expression.
-    // However, having clause is parsed after group by, and currently, it is only
-    // supported with alias expression.
-
-    if (!postgresQueryParser.getPgSelections().containsKey(key)) {
-      throw new UnsupportedOperationException(
-          "Having clause is only supported with alias expression"
-              + "for aggregation and functional expression");
+    if (postgresQueryParser.getPgSelections().containsKey(key)) {
```
kotharironak, line 35: Removed the restriction added to move on to allow only functional expressions as part of having clause.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -102,6 +110,28 @@ public static String prepareCast(String field, Type type) {
     }
   }
 
+  public static String prepareCastForFieldAccessor(String field, Object value) {
```
kotharironak, line 113: We are using the field accessor of jsonb ethe lement in `group by` clause. So, it needs to same in  `having` clause. So, for string type, we need different casting.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -303,7 +333,7 @@ public static List<String> splitNestedField(String nestedFieldName) {
   }
 
   public static boolean isEncodedNestedField(String fieldName) {
-    return fieldName.contains(DOT_STR) ? true : false;
+    return fieldName.contains(DOT_STR) || fieldName.contains(DOT) ? true : false;
```
kotharironak, line 336: Like mongo, support alias with `.` as the hierarchy of fields. See the test case `testAggregateWithNestedFields` in `DocStoreQueryV1Test.java`

PR #114, opened 2022-08-09
fix: handle existing document gracefully

kotharironak:


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -730,7 +730,9 @@ private Optional<Long> getCreatedTime(Key key) throws IOException {
     CloseableIterator<Document> iterator = searchDocsForKeys(Set.of(key));
     if (iterator.hasNext()) {
       JsonNode existingDocument = getDocAsJSON(iterator.next());
-      return Optional.of(existingDocument.get(DocStoreConstants.CREATED_TIME).asLong());
+      if (existingDocument.has(DocStoreConstants.CREATED_TIME)) {
+        return Optional.of(existingDocument.get(DocStoreConstants.CREATED_TIME).asLong());
+      }
```
suresh-prakash, line 735: Nit: Can be simplified as 
```suggestion
      return Optional.ofNullable(existingDocument.get(DocStoreConstants.CREATED_TIME)).map(JsonNode::asLong);
```

suresh-prakash (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: Let me handle it follow up, not sure where the exception was - `existingDocument.has(DocStoreConstants.CREATED_TIME)` or later.

PR #118, opened 2022-08-17
Atomic read and update API with implementation

suresh-prakash:
## Description
Currently, we have an API to atomically update (based on an id and a filter). This PR adds the API to atomically update the first document matching the filter criteria without passing in the id.

### Testing
Added unit and integration tests

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -378,6 +378,14 @@ public CloseableIterator<Document> aggregate(
     return executeQueryV1(query);
   }
 
+  @Override
+  public Optional<Document> atomicReadAndUpdateDocument(
+      final org.hypertrace.core.documentstore.query.Query query, final Document updateDocument)
+      throws IOException {
+    // TODO: Implement
+    throw new UnsupportedOperationException(UNSUPPORTED_QUERY_OPERATION);
```
kotharironak, line 386: Let's have a basic implementation of it on Postgres's side too. we need parity before the API is getting used.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -80,6 +81,24 @@ public interface Collection {
...
+   * @return The document before updating if exists, otherwise an empty optional
+   * @throws IOException if there was any error in updating/fetching the document
+   * @implSpec The definition of an update here is
+   *     <ol>
+   *       <li>The existing fields/sub-documents will be updated
+   *       <li>New fields/sub-documents will be created if they do not exist
+   *       <li>None of the existing fields/sub-documents will be removed
+   *     </ol>
+   */
+  Optional<Document> atomicReadAndUpdateDocument(
```
kotharironak, line 98: if we could align name with other APIs - `updateAndReturnFirst` or `updateAndReturn`?  Or `update` should be fine?

we have that has a similar use case.
```UpdateResult update(Key key, Document document, Filter condition) throws IOException;```

So, this can be :
```update(final org.hypertrace.core.documentstore.query.Query query, final Document updateDocument) throws IOException;```

File: document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java [hunk 3]
```diff
@@ -182,5 +184,31 @@ CloseableIterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Document>
...
+   * @return The document <strong>before updating</strong> if exists, otherwise an empty optional
+   * @throws IOException if there was any error in updating/fetching the document
+   * @implSpec The definition of an update here is
+   *     <ol>
+   *       <li>The existing sub-documents will be updated
+   *       <li>New sub-documents will be created if they do not exist
+   *       <li>None of the existing sub-documents will be removed
+   *     </ol>
+   */
+  Optional<Document> update(
```
kotharironak, line 208: This will update a single document only, right? But, can modify multiple child fields of the matched document?

suresh-prakash: > This will update a single document only, right?

Yes

> But, can modify multiple child fields of the matched document?

Yes

document-store/src/integrationTest/resources/mongo/updatable_collection_data.json
```diff
@@ -0,0 +1,175 @@
+[
+  {
+    "_id": 1,
+    "item": "Soap",
+    "price": 10,
+    "quantity": 2,
+    "date": "2014-03-01T08:00:00Z",
```
kotharironak, line 7: As the query looks for `< 2022-08-09T18:53:17Z`, we have no document beyond it. So, can we add one document with data - `2023-08-09T18:53:17Z` for completeness?

suresh-prakash: Done.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -388,6 +399,52 @@ public CloseableIterator<Document> aggregate(
     return executeQueryV1(query);
   }
 
+  @Override
+  public Optional<Document> update(
+      final org.hypertrace.core.documentstore.query.Query query,
+      final java.util.Collection<SubDocumentUpdate> updates)
+      throws IOException {
+
+    try (final Connection connection = client.getNewConnection()) {
```
kotharironak, line 408: We are using a fixed set of static connections right now. And so, we are reusing the same connection till it is valid. I think,  if we can reuse the same connecction like any other API - getConnection() and reset connection back to `connection.setAutoCommit(true)` at the end. That should work fine, what do you think?
cc: @saxenakshitiz

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -388,6 +399,52 @@ public CloseableIterator<Document> aggregate(
...
+      connection.setAutoCommit(false);
+
+      org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser parser =
+          new org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser(
+              collectionName, query);
+      final String selectQuery = queryBuilder.buildSelectQueryForUpdateSkippingLocked(parser);
+
+      try (final PreparedStatement pStmt = connection.prepareStatement(selectQuery)) {
```
kotharironak, line 416: Q: Does this final query implicitly put `LIMIT 1` clause if it is not present in client call of the main query request?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresQueryBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+
+@RequiredArgsConstructor
+public class PostgresQueryBuilder {
+  private final String collectionName;
+  private final PostgresSubDocumentValueParser subDocValueParser =
+      new PostgresSubDocumentValueParser();
+
+  public String buildSelectQueryForUpdateSkippingLocked(
+      org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser parser) {
+    final String selections = parser.getSelectionsWithImplicitId();
```
kotharironak, line 19: We are not expecting a query with `unnest expression` for update API right?
If, required, let's add it when needed.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
```diff
@@ -62,76 +62,68 @@ public String parse() {
...
+  public String getSelectionsWithImplicitId() {
+    return getSelections() + ", " + String.format("%s AS %s", ID, IMPLICIT_ID_ALIAS);
+  }
+
+  public Optional<String> parseSelection() {
     return Optional.ofNullable(PostgresSelectTypeExpressionVisitor.getSelections(this));
   }
 
-  private Optional<String> parseFilter() {
+  public Optional<String> parseFilter() {
```
kotharironak, line 102: Currently, the parser is a stateful and parse thing in a specific order. So, not sure we should go it this way making independent clause methods public. Instead, we can add another parse method as per our need?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/subdoc/PostgresSubDocumentUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+      pStatement.setString(3, id);
+
+      pStatement.executeUpdate();
+    }
+  }
+
+  public void updateLastUpdatedTime(final Connection connection, final String id)
+      throws SQLException {
+    final SubDocumentUpdate lastUpdatedTimeUpdate =
+        SubDocumentUpdate.of(SubDocument.implicitUpdatedTime(), clock.millis());
```
kotharironak, line 43: As we need milliseconds only, can we use `System.currentTimeMillis()`?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -74,6 +75,7 @@ public class MongoCollection implements Collection {
 
   private final com.mongodb.client.MongoCollection<BasicDBObject> collection;
   private final MongoQueryExecutor queryExecutor;
+  private final Clock clock;
```
kotharironak, line 78: I think, we can directly use `System.currentTimeMillis()`.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -475,6 +466,44 @@ public CloseableIterator<Document> aggregate(
...
 
+  @Override
+  public Optional<Document> update(
+      final org.hypertrace.core.documentstore.query.Query query,
+      final java.util.Collection<SubDocumentUpdate> updates)
+      throws IOException {
+    try {
+      final BasicDBObject selections = getSelections(query);
+      final BasicDBObject sorts = getOrders(query);
+      final FindOneAndUpdateOptions options = new FindOneAndUpdateOptions();
```
kotharironak, line 477: nit: for readability, we can pass the argument `ReturnDocument.BEFORE` or put the comment that it is the default.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
```diff
@@ -52,6 +52,11 @@ public synchronized Connection getConnection() {
     return connection;
   }
 
+  public Connection getNewConnection() throws SQLException {
```
kotharironak, line 55: Check the comment on this.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
suresh-prakash: If we use the same connection and keep calling the `setAutoCommit()`, then the **updates can be lost** (since we don't control concurrency in this library).

1. Thread 1 calls the new `update` method and we disable auto-commit for the shared connection
2. Thread 1 completes the operation and commits
3. Thread 1 is interrupted (before enabling auto-commit). Thread 2 calls the existing `upsert` method and (since we haven't yet enabled auto-commit and the commit is already complete), this upsert would be lost
4. Thread 1 resumes and enables the auto-commit

Another issue I see is incorrect aborting. Consider the scenario.
1. Thread 1 calls the new `update` method and we disable auto-commit for the shared connection
2. Thread 1 is interrupted. Thread 2 calls the existing `upsert` method (and is not committed)
3. Thread 1 resumes, but fails. We abort the transaction (including the `upsert` done by Thread 2).


A proper solution would be to use a connection pool. One could consider using two connections: one with auto-commit on and the other with auto-commit off. But, that wouldn't scale either because.
1. Thread 1 calls the new `update` method
2. Thread 2 calls the new `update` method
3. Thread 1 commits (at a point where thread 2 has not completed executing).

The same issue will persist with abort as well.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
suresh-prakash: Yes. Refer. line number 28 in org.hypertrace.core.documentstore.postgres.PostgresQueryBuilder.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresQueryBuilder.java
suresh-prakash: That's right.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParser.java
suresh-prakash: Makes sense. Will merge `PostgresQueryBuilder` with this class.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/subdoc/PostgresSubDocumentUpdater.java
suresh-prakash: Preferring `Clock` instead of `System.currentTimeMillis()` because of the improved [testability](https://www.baeldung.com/java-override-system-time).

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
suresh-prakash: Preferring `Clock` instead of `System.currentTimeMillis()` because of the improved [testability](https://www.baeldung.com/java-override-system-time).

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
suresh-prakash: Sure. Will do.

kotharironak (approved):
Let's work on the connection pool as a separate PR. Can you create a ticket and either assign it to me or you?

LGTM

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
kotharironak: As discussed, lets see if we can use a separate connection pool for APIs which need explicit commit.

PR #120, opened 2022-08-23
Update Postgres Filter Type Expression Parser to make it extendible to non-identifier type LHS

suresh-prakash:
## Description
Currently, filters only support identifier LHS and constant RHS. We may want to extend it to support functional expression on the LHS. This PR is refactoring the existing code so that the Postgres filter type expression parser could be extended later to support functional expressions.

### Testing
All the existing unit and integration tests are passing.
No new test case is required because, it is just refactoring.

### Checklist:

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
@@ -52,17 +52,18 @@ public class PostgresFilterTypeExpressionVisitor implements FilterTypeExpression
     PostgresSelectTypeExpressionVisitor lhsVisitor =
         new PostgresDataAccessorIdentifierExpressionVisitor(postgresQueryParser, getType(value));
 
-    String lhsExpression = lhs.accept(lhsVisitor);
+    final String fieldName = lhs.accept(identifierVisitor);
+    final String lhsExpression = lhs.accept(lhsVisitor);
 
     FieldToPgColumn fieldToPgColumn =
-        postgresQueryParser.getToPgColumnTransformer().transform(lhsExpression);
+        postgresQueryParser.getToPgColumnTransformer().transform(fieldName);
 
     if (fieldToPgColumn.getTransformedField() == null)
       throw new UnsupportedOperationException("jsonb types in where clause is not yet supported");
 
     return PostgresUtils.parseNonCompositeFilter(
-        lhs.accept(identifierVisitor),
         fieldToPgColumn.getTransformedField(),
+        lhsExpression,
         fieldToPgColumn.getPgColumn(),
         operator.toString(),
         value,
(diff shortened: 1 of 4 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -43,19 +45,23 @@ public String visit(final RelationalExpression expression) {
...
     // Only an identifier LHS and a constant RHS is supported as of now.
-    PostgresSelectTypeExpressionVisitor lhsVisitor = new PostgresIdentifierExpressionVisitor();
     PostgresSelectTypeExpressionVisitor rhsVisitor = new PostgresConstantExpressionVisitor();
-
-    String fieldName = lhs.accept(lhsVisitor);
     Object value = rhs.accept(rhsVisitor);
+    PostgresIdentifierExpressionVisitor identifierVisitor =
+        new PostgresIdentifierExpressionVisitor();
+    PostgresSelectTypeExpressionVisitor lhsVisitor =
+        new PostgresDataAccessorIdentifierExpressionVisitor(postgresQueryParser, getType(value));
```
kotharironak, line 53: I think we need `PostgresFunctionExpressionVisitor` to support it, right?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -43,19 +45,23 @@ public String visit(final RelationalExpression expression) {
...
+
+    String lhsExpression = lhs.accept(lhsVisitor);
 
     FieldToPgColumn fieldToPgColumn =
-        postgresQueryParser.getToPgColumnTransformer().transform(fieldName);
+        postgresQueryParser.getToPgColumnTransformer().transform(lhsExpression);
```
kotharironak, line 58: This will also a problem as we need orgFieldName while searching the map, not with dataAccessor pattern

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
suresh-prakash: Yes. This will come in a follow-up PR along with unit & integration tests.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
suresh-prakash: Oh! Yes. Will fix it.

kotharironak (approved):
lgtm

PR #119, opened 2022-08-19
feat: add connection pooling for postgres connections

saxenakshitiz:

PR #121, opened 2022-08-24
Revert "feat: add connection pooling for postgres connections"

saxenakshitiz:
Reverts hypertrace/document-store#119


kotharironak:
The reason we are reverting is that the above solution doesn't suffice our requirement. We enriched data during ingestion pipeline, and in the process, it talks to entity service which internally fetches data via document-store. This needs a sort of live connection and decoupling from `PreparedStatement & ResultSet`. Currently, in the `PoolableConnection`, `resultset` is coupled with the borrowed connection. So, if we return back the connection to pull early before data is read, it also closes the underlying result set.

So, we are considering a different approach to re-establishing a connection if it no longer remains valid.

kotharironak (approved):

PR #122, opened 2022-08-24
feat: create new connection on connection validation error in postgres

saxenakshitiz:


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
```diff
@@ -0,0 +1,112 @@
...
+        close();
+        newConnection();
+      }
+    } catch (SQLException sqle) {
+      throw new RuntimeException(sqle);
+    }
+    return connection;
+  }
+
+  private synchronized boolean isConnectionValid(Connection connection) {
```
kotharironak, line 55: Do we need  `synchronized`  on `private` method? I think we should be fine without it as at any point in the time only one public entry to the getConnection function. what do you think?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
```diff
@@ -0,0 +1,112 @@
...
+      log.debug("Unable to check if the underlying connection is valid", sqle);
+      return false;
+    }
+  }
+
+  private synchronized void newConnection() throws SQLException {
+    int attempts = 0;
+    while (attempts < maxConnectionAttempts) {
+      try {
+        ++count;
```
kotharironak, line 75: I think, If, we want to know how many different new connections are used throughout, we should increment the count in `getConnection` after line no: `47`. 

So, between old valid connection = 2, and newValidConenction=3, there were X attempts were made. This information is logged. This will also help in debugging actually which connection was closed, it's kind of id of that connection.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
saxenakshitiz: Just a symbolic one. One the public one is synchronized, it actually ensures all subsequent to private methods are synchronized as well. Just wanted to call it out.

kotharironak: wouldn't it aquire its own lock? would it add to additional cost?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
saxenakshitiz: FIxed

kotharironak (approved):

PR #128, opened 2022-09-19
chore: rename DISTINCT aggregation operator as DISTINCT_ARRAY

SrikarMannepalli:
## Description
This PR renames DISTINCT aggregation operator to  DISTINCT_ARRAY. 
 

### Testing
Updated all the unit tests accordingly and verified that all pass.

### Checklist:

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
@@ -3,7 +3,7 @@ package org.hypertrace.core.documentstore.expression.operators;
 public enum AggregationOperator {
   AVG,
   COUNT,
-  DISTINCT, // This operator generates an array of distinct values
+  DISTINCT_ARRAY, // This operator generates an array of distinct values
   DISTINCT_COUNT,
   SUM,
   MIN,
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -2,7 +2,7 @@ package org.hypertrace.core.documentstore;
 
 import static org.hypertrace.core.documentstore.expression.operators.AggregationOperator.AVG;
 import static org.hypertrace.core.documentstore.expression.operators.AggregationOperator.COUNT;
-import static org.hypertrace.core.documentstore.expression.operators.AggregationOperator.DISTINCT;
+import static org.hypertrace.core.documentstore.expression.operators.AggregationOperator.DISTINCT_ARRAY;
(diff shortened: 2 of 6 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
```diff
@@ -4,6 +4,7 @@ public enum AggregationOperator {
   AVG,
   COUNT,
   DISTINCT, // This operator generates an array of distinct values
+  DISTINCT_ARRAY, // This operator generates an array of distinct values
```
suresh-prakash, line 7: The comment in the existing code (`This operator generates an array of distinct values`) is a bit misleading.
DISTINCT can be used with normal fields or array fields (after unwinding them).

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
```diff
@@ -4,6 +4,7 @@ public enum AggregationOperator {
   AVG,
   COUNT,
   DISTINCT, // This operator generates an array of distinct values
```
suresh-prakash, line 6: Can we add a comment here that this operator (1) has been deprecated and (2) will be removed, along with the (3) alternate operator being `DISTINCT_COUNT`?
CC: @kotharironak

suresh-prakash: In fact, we can get rid of the old one in favour of the new operator since it is a library.

suresh-prakash (approved):

suresh-prakash:
+1 for meticulously updating the PR title and description!

document-store/src/main/java/org/hypertrace/core/documentstore/expression/operators/AggregationOperator.java
aaron-steinfeld: Sorry late on the review, but libraries have the same deprecation concerns as services. If it's not a lot of work, we should restore the old one and deprecate. The alternative is going make sure every consumer of the lib is up to date, which is likely more work.

aaron-steinfeld: To elaborate a little bit, it's not quite the same - it takes two different forms
- compile time breakages on version upgrade. You don't want that from a library, but for an internal library this is less of a concern. It's better than runtime breakages which you can often see from a break of a service contract.
- transitives and version resolution. This is the real gotcha - If two different modules are using two different versions of doc store but running in the same VM, gradle resolves to the later version. The one that compiled against an earlier version would lead to a runtime failure. That's why java lib APIs always have the same backwards compatibility requirements as services, so we can resolve multiple libs in the same classloader depending on different versions.

SrikarMannepalli: Oh okay. Understood. Will restore DISTINCT.

PR #133, opened 2022-09-29
Filter by key support

suresh-prakash:
## Description
Previously, there was no direct way to filter by keys in v1 Query. The only, roundabout way was to use `_id` field, but filtering by this internal field will not use index for Postgres since `id` is an external field (outside of the document).
This PR addresses this by introducing a KeyExpression that can be filtered on.

### Testing
Added unit and integration tests for both Postgres and Mongo

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/KeyExpression.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Override
+  public <T> T accept(final FilterTypeExpressionVisitor visitor) {
+    return visitor.visit(this);
+  }
+
+  @Override
+  public String toString() {
+    return IMPLICIT_ID + " = " + StringUtils.wrap(key.toString(), "'");
```
skjindal93, line 31: Do we use `StringUtils.wrap` to avoid injection attacks? In other words, can `key` be maliciously crafted, since we are doing a basic string concatenation and not using string builders, or is it taken care of in the pipeline later?

Or, is `toString` method being used mostly for logging purposes?

suresh-prakash: >  is toString method being used mostly for logging purposes?

That's correct. `toString()` is only for logging (except the FunctionalExpression.toString() in the case of MongoDB, which is escaped during the parsing step). We are using PreparedStatement in `PostgresFilterTypeExpressionVisitor.visit(KeyExpression)` during parsing.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/Utils.java
```diff
@@ -99,14 +100,17 @@ public static Map<String, Object> convertDocumentToMap(Document document)
...
+      throws IOException {
     String fileContent = readFileFromResource(filePath).orElseThrow();
     List<Map<String, Object>> expected = convertJsonToMap(fileContent);
 
     int actualSize = 0;
     List<Map<String, Object>> actual = new ArrayList<>();
     while (documents.hasNext()) {
-      actual.add(convertDocumentToMap(documents.next()));
+      final Map<String, Object> doc = convertDocumentToMap(documents.next());
+      removeDateRelatedFields(dataStoreName, doc);
```
kotharironak, line 112: Was this not a problem on the existing (or prior) test? or is this one not used everywhere?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -1235,15 +1245,147 @@ public void testFindWithSortingAndPagination(String datastoreName) throws IOExce
...
+  @ParameterizedTest
+  @MethodSource("databaseContextBoth")
+  public void testFindWithKeyAndMatchingRelationalFilter(final String datastoreName)
+      throws IOException {
+    final Collection collection = getCollection(datastoreName);
+    final org.hypertrace.core.documentstore.query.Filter filter =
+        org.hypertrace.core.documentstore.query.Filter.builder()
+            .expression(
+                and(
+                    KeyExpression.of(new SingleValueKey(TENANT_ID, "7")),
```
kotharironak, line 1341: **Just a note:** Being a document store, which allows storing data as (key, document), we can have query language for `key` and `document` parts separately. Here, we are taking a single Query language approach for both the fields - key and document. So the query can have both key and document-related expressions as shown here in the example.
This adds the functionality of selecting certain documents based on a set of IDs, and further filtering document values.

kotharironak: Nit: Could be better if the test cases use multiple IDs as `OR` and further filter.

kotharironak (approved):
Minor comments, LGTM

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/utils/Utils.java
suresh-prakash: This was not used everywhere because this is required only when the selection list is empty.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
suresh-prakash: > Here, we are taking a single Query language approach for both the fields - key and document. So the query can have both key and document-related expressions as shown here in the example.
This adds the functionality of selecting certain documents based on a set of IDs, and further filtering document values.

Yes. This is done intentionally so that we have can combine key and other fields in querying making the query language more expressive. For example, we have `delete(Key)` and `delete(Filter)`. Tomorrow if we have a use-case to delete a specific document only when a condition is satisfied (like delete the document with id 7 only if the last updated time is older than 30 days), then we can neither use `delete(Key)` nor `delete(Filter)`.

Making the `Key` a part of the `Filter` not only allows for such use-cases, but also helps in merging many methods.

suresh-prakash: > Could be better if the test cases use multiple IDs as OR and further filter.

Sure, will add in a separate PR.

PR #138, opened 2022-10-14
Fix Mongo for functional LHS expression in the relational filters

suresh-prakash:
## Description
Previously functional LHS was handled in the aggregation pipeline by adding extra fields via the '$addFields' stage. This limits the `find()` and the new atomic `update()` APIs from using functional LHS in the filters. This PR addresses this issue by getting rid of the `$addFields` pipeline stage and using the `$expr` instead.

### Testing
Added an integration test.

### Checklist:


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -626,6 +626,34 @@ public void testQueryWithFunctionalLeftHandSideInFilter(final String dataStoreNa
...
+                        .operand(IdentifierExpression.of("price"))
+                        .build(),
+                    GT,
+                    ConstantExpression.of(50)))
+            .addSort(IdentifierExpression.of("item"), DESC)
+            .build();
+
+    final Iterator<Document> resultDocs = collection.find(query);
+    assertDocsAndSizeEqual(
+        dataStoreName, resultDocs, "query/test_aggr_functional_lhs_in_filter_response.json", 3);
```
kotharironak, line 653: nit (upto you): shall we remove `aggr` from file name `test_aggr_functional_lhs_in_filter_response.json` -> `test_functional_lhs_in_filter_response.json` as it is used now both with `find` API and `aggregate` API ?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoQueryExecutor.java
```diff
@@ -80,7 +79,14 @@ public MongoCursor<BasicDBObject> aggregate(final Query originalQuery) {
             .collect(Collectors.toList());
 
     logPipeline(pipeline);
-    AggregateIterable<BasicDBObject> iterable = collection.aggregate(pipeline);
+
+    AggregateIterable<BasicDBObject> iterable;
```
kotharironak, line 83: nit: As iterable is not used in the exception block, shall we do as below so if it's not null?
```
try {
      AggregateIterable<BasicDBObject> iterable = collection.aggregate(pipeline);
      return iterable.cursor();
    } catch (final MongoCommandException e) {
      log.error("Execution failed for query: {}. Aggregation Pipeline: {}", query, pipeline);
      throw e;
    }
 ```

kotharironak (commented):
Rest lgtm

kotharironak (approved):
Have minor comments, lets address them as follow up.

kotharironak (approved):

PR #142, opened 2022-11-05
fix: update the like operator to handle posix expressions

saxenakshitiz:


suresh-prakash (commented):
Changes look good. Could you please add an integration test for covering both Postgres and MongoDB?

suresh-prakash (approved):

PR #143, opened 2022-11-10
fix: fix conversion of outer column values to json node

saxenakshitiz:


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -349,6 +350,178 @@ public void testFindWithNestedFields(String dataStoreName) throws IOException {
     testCountApi(dataStoreName, query, "query/filter_on_nested_fields_response.json");
   }
 
+  @ParameterizedTest
+  @MethodSource("databaseContextPostgres")
+  void testAggregateWithId(String dataStoreName) throws IOException {
```
suresh-prakash, line 355: Could you re-use the existing `collection_data.json` to put the additional fields instead of creating a separate document for each test?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -1190,6 +1215,25 @@ private void handleNestedField(
...
+      case INTEGER:
+        return IntNode.valueOf(Integer.valueOf(columnValue));
+      case BOOLEAN:
+        return BooleanNode.valueOf(Boolean.valueOf(columnValue));
+      case DOUBLE:
+        return DoubleNode.valueOf(Double.valueOf(columnValue));
+      case FLOAT:
+        return FloatNode.valueOf(Float.valueOf(columnValue));
+      default:
+        return null;
```
suresh-prakash, line 1233: Can we prefer returning Optional instead of null?
That way, the callers would be forced to check for the presence.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -349,6 +350,178 @@ public void testFindWithNestedFields(String dataStoreName) throws IOException {
     testCountApi(dataStoreName, query, "query/filter_on_nested_fields_response.json");
   }
 
+  @ParameterizedTest
+  @MethodSource("databaseContextPostgres")
```
suresh-prakash, line 354: All the tests should be for both MongoDB and Postgres so that we have compatibility.
```suggestion
  @MethodSource("databaseContextBoth")
```

saxenakshitiz: The id being returned in case of mongo is different from what is being returned from postgres. I had discussion with @kotharironak, and he said it will be fixed later.

suresh-prakash: Sure.

suresh-prakash (approved):

PR #145, opened 2022-12-02
Update operation support

suresh-prakash:
## Description
Previously only the "SET" operation was supported on the update, which can only assign a value to a field.
This PR aims at the other update operations including
1. UNSET - Remove the specified path from the document
2. ADD_TO_LIST_IF_ABSENT - Add all (single or multiple) the non-existing values to an array
3. APPEND_TO_LIST - Add all (single or multiple) the values to an array
4. REMOVE_ALL_FROM_LIST - Removes all occurrences of the given elements (single or multiple) from an array

Additional changes:
- Make KeyExpression work with multiple keys at once
- Introduce common validator for MongoDB and Postgres

Note: These operations (except APPEND_TO_LIST) were already supported in the form of array operations and delete sub-document. This PR implements them for the new `update` APIs so that the execution is truly atomic.

### Testing
- Added integration tests for both MongoDB and Postgres
- Added unit tests for the new validator implementation

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/model/subdoc/UpdateOperator.java
```diff
@@ -0,0 +1,9 @@
+package org.hypertrace.core.documentstore.model.subdoc;
+
+public enum UpdateOperator {
+  SET,
+  UNSET,
```
kotharironak, line 5: What does `UNSET` mean? Does it remove the full sub-document path?

suresh-prakash: Yes.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -1977,8 +1982,201 @@ public void testAtomicUpdateDocumentWithoutSelections(final String datastoreName
...
+
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testRemoveAllOccurrancesFromIntegerList(final String datastoreName) throws IOException {
+      final Collection collection = getCollection(datastoreName, UPDATABLE_COLLECTION_NAME);
+      createCollectionData("query/updatable_collection_data.json", UPDATABLE_COLLECTION_NAME);
+
+      final SubDocumentUpdate unset =
+          SubDocumentUpdate.builder().subDocument("sales").operator(UNSET).build();
+      final SubDocumentUpdate add =
```
kotharironak, line 2086: Is the default operator `SET`? Can we explicitly provide the operator? Also nit: `add` -> `set`

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -1977,8 +1982,201 @@ public void testAtomicUpdateDocumentWithoutSelections(final String datastoreName
...
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testAddToListIfAbsentDoesNotDeduplicateTheExistingList(final String datastoreName)
+        throws IOException {
+      final Collection collection = getCollection(datastoreName, UPDATABLE_COLLECTION_NAME);
+      createCollectionData("query/updatable_collection_data.json", UPDATABLE_COLLECTION_NAME);
+
+      final SubDocumentUpdate unset =
```
kotharironak, line 2124: Do we need `unset` on `sales` for the tests we are doing - or is this just to reduce JSON file size? I see the same condition in the previous test too so was wondering - is it required?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -1977,8 +1982,201 @@ public void testAtomicUpdateDocumentWithoutSelections(final String datastoreName
...
+              .build();
+
+      final Query query = Query.builder().build();
+      final List<SubDocumentUpdate> updates = List.of(add, unset);
+
+      assertThrows(
+          IOException.class,
+          () ->
+              collection.bulkUpdate(
+                  query, updates, UpdateOptions.builder().returnDocumentType(NONE).build()));
```
kotharironak, line 2178: Why should it throw an exception?
- First, we are creating the property - `props.added.list`
- Second, its removes `props.added`
So, what is not allowed - a partial path?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
suresh-prakash: > Is the default operator SET?

Yes.

> Can we explicitly provide the operator?

Sure. Will update.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
suresh-prakash: > is this just to reduce JSON file size?

Yes. That's the idea. Otherwise not required.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
suresh-prakash: If we allow such paths, we need to ensure that MongoDB and Postgres run those updates in the same order.
Basically, the result would become arbitrary breaking the compatibility between MongoDB and Postgres.

In other words, this is to let the clients know that the updates can be executed in any order, which is why it takes in a `Collection` rather than a list.

document-store/src/main/java/org/hypertrace/core/documentstore/model/subdoc/SubDocumentUpdate.java
```diff
@@ -36,7 +46,7 @@ public static SubDocumentUpdate of(final String subDocumentPath, final Boolean[]
   }
 
   public static SubDocumentUpdate of(final String subDocumentPath, final SubDocumentValue value) {
-    return new SubDocumentUpdate(SubDocument.builder().path(subDocumentPath).build(), value);
+    return new SubDocumentUpdate(SubDocument.builder().path(subDocumentPath).build(), SET, value);
```
kotharironak, line 49: Why are we making the `SET` default? For backward compatibility?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFilterTypeExpressionParser.java
```diff
@@ -32,7 +33,9 @@ public Map<String, Object> visit(final RelationalExpression expression) {
   @SuppressWarnings("unchecked")
   @Override
   public Map<String, Object> visit(final KeyExpression expression) {
-    return Map.of(ID_KEY, expression.getKey().toString());
+    return Map.of(
+        ID_KEY,
+        Map.of("$in", expression.getKeys().stream().map(Key::toString).toArray(String[]::new)));
```
kotharironak, line 38: I think we will have to check the performance of `in` and `eq` for a single key.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/update/parser/MongoPullAllOperationParser.java
```diff
@@ -0,0 +1,24 @@
...
+import static org.hypertrace.core.documentstore.model.subdoc.UpdateOperator.REMOVE_ALL_FROM_LIST;
+import static org.hypertrace.core.documentstore.mongo.MongoUtils.merge;
+
+import com.mongodb.BasicDBObject;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.hypertrace.core.documentstore.model.subdoc.UpdateOperator;
+
+@AllArgsConstructor
+public class MongoPullAllOperationParser extends MongoOperationParser {
```
kotharironak, line 12: nit/thought: `Pull` is internal to Mongo. So, if we can keep the name as our Update Operator, we may able to map them easily while reading.
 e.g naming as `MongoRemoveAllFromListOperationParser`

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/update/parser/MongoPushOperationParser.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.hypertrace.core.documentstore.model.subdoc.UpdateOperator.APPEND_TO_LIST;
+import static org.hypertrace.core.documentstore.mongo.MongoUtils.merge;
+
+import com.mongodb.BasicDBObject;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.hypertrace.core.documentstore.model.subdoc.UpdateOperator;
+
+@AllArgsConstructor
+public class MongoPushOperationParser extends MongoOperationParser {
```
kotharironak, line 12: nit: `MongoPushOperationParser` -> `MongoAppendToListOperationParser`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -66,7 +68,10 @@ public String visit(final RelationalExpression expression) {
   @Override
   public String visit(final KeyExpression expression) {
     return PostgresUtils.prepareParsedNonCompositeFilter(
-        ID, EQ.name(), expression.getKey().toString(), postgresQueryParser.getParamsBuilder());
+        ID,
```
kotharironak, line 71: Same comment on perf above. I think, if the size of the keys is one, we should use `EQ` operator.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
suresh-prakash: To elaborate a bit on your example, 
If SET `props.added.list` gets executed first, and then UNSET `props.added`, the result would be without the `props.added` path.
However, if we reverse the order, then `props.added.list` would be set. 
Basically, it creates confusion depending on the execution order.

If they clients really want such an update, they will have to invoke the `update` API twice once for each path one after the other so that they have clarity on the execution order.

document-store/src/main/java/org/hypertrace/core/documentstore/commons/CommonUpdateValidator.java
```diff
@@ -0,0 +1,45 @@
...
+  private void validateUpdateNotPerformedInSameHierarchy(Collection<SubDocumentUpdate> updates)
+      throws IOException {
+    final String[] sortedPaths =
+        updates.stream()
+            .map(SubDocumentUpdate::getSubDocument)
+            .map(SubDocument::getPath)
+            .sorted()
+            .toArray(String[]::new);
+    for (int i = 1; i < sortedPaths.length; i++) {
+      if (arePartOfSameHierarchy(sortedPaths[i - 1], sortedPaths[i])) {
```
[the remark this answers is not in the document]
suresh-prakash: Please refer to my comment here: https://github.com/hypertrace/document-store/pull/145#discussion_r1042108672

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/KeyExpression.java
```diff
@@ -14,22 +20,29 @@
 @Value
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class KeyExpression implements FilterTypeExpression {
-  Key key;
+  List<Key> keys;
```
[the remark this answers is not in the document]
suresh-prakash: >  Is this to support IN expression?

Yes. Basically, when we want to filter by multiple keys.

> will always apply IN?

Yes. We always do `IN`.

document-store/src/main/java/org/hypertrace/core/documentstore/model/subdoc/SubDocumentUpdate.java
suresh-prakash: Yup. For backward compatibility.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFilterTypeExpressionParser.java
suresh-prakash: Reverted to use `eq` for a single key.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/KeyExpression.java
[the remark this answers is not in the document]
suresh-prakash: Reverted the code to use `eq` for a single key.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/update/parser/MongoPullAllOperationParser.java
suresh-prakash: Sure.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
suresh-prakash: Sure. Updated.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -282,18 +285,21 @@ Optional<Document> update(
...
-   *     empty
+   * @throws IOException if there was any error in updating/fetching the documents or argument
+   *     validation failed (like no updates passed)
    * @implSpec The definition of an update here is
    *     <ol>
    *       <li>The existing sub-documents will be updated
    *       <li>New sub-documents will be created if they do not exist
    *       <li>None of the existing sub-documents will be removed
    *     </ol>
+   *     <p>If more than one update is modifying the same hierarchy (e.g.: a->b and a->b->c), should
```
kotharironak, line 300: I think, it's up to the client, isn't it? We can discuss this offline, I am not fully clear here. For now, this is not a blocker.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -60,18 +61,24 @@ public String visit(final RelationalExpression expression) {
...
 
   @SuppressWarnings("unchecked")
   @Override
   public String visit(final KeyExpression expression) {
-    return PostgresUtils.prepareParsedNonCompositeFilter(
-        ID,
-        IN.name(),
-        expression.getKeys().stream().map(Key::toString).collect(toUnmodifiableList()),
-        postgresQueryParser.getParamsBuilder());
+    return expression.getKeys().size() == 1
```
kotharironak, line 71: Thanks for taking care of this.

kotharironak (approved):

PR #153, opened 2023-06-08
Support for increment update query operator

shantanu-vsbhosale:
## Description
Support for `increment` update operator. This is required to update the numeric value in db itself rather than incrementing it in-memory.

### Testing
Added integration tests for increment update operator.

### Checklist:


document-store/src/main/java/org/hypertrace/core/documentstore/model/subdoc/UpdateOperator.java
```diff
@@ -6,4 +6,5 @@ public enum UpdateOperator {
   REMOVE_ALL_FROM_LIST,
   ADD_TO_LIST_IF_ABSENT,
   APPEND_TO_LIST,
+  INCREMENT
```
suresh-prakash, line 9: Still think if we name this as `ADD`, 
1. it can be extended to adding the values of 2 fields and stored in a 3rd one
2. pairity operators like `MULTIPLY` can be introduced in the future

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/update/parser/PostgresIncrementValueParser.java
```diff
@@ -0,0 +1,26 @@
...
+  @Override
+  public String parseLeaf(UpdateParserInput input) {
+    final Params.Builder paramsBuilder = input.getParamsBuilder();
+    final PostgresSubDocumentValueParser valueParser =
+        new PostgresSubDocumentValueParser(paramsBuilder);
+
+    paramsBuilder.addObjectParam(formatSubDocPath(input.getPath()[0]));
+    final String parsedValue = input.getUpdate().getSubDocumentValue().accept(valueParser);
+    return String.format(
+        "jsonb_set(%s, ?::text[], (COALESCE(%s->>'%s', '0')::float + %s::float)::text::jsonb)",
```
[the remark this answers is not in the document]
suresh-prakash: Actually, this seems to be a bigger effort. Let's check the behavior of both databases. If they are the same (either throwing an error or considering a 0 value), let's just add an integration test to verify that.
That way, even if a third database comes in the future, it would be forced to adhere to the same behavior.

In case both Postgres and Mongo are behaving differently, we don't have a way but to add the explicit check.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/update/parser/MongoIncrementOperationParser.java
```diff
@@ -0,0 +1,24 @@
...
+public class MongoIncrementOperationParser extends MongoOperationParser {
+  private static final String INCREMENT_CLAUSE = "$inc";
+
+  @Override
+  UpdateOperator operator() {
+    return INCREMENT;
+  }
+
+  @Override
+  BasicDBObject wrapWithOperator(List<BasicDBObject> parsed) {
```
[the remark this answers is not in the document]
shantanu-vsbhosale: TBH, I am not clear what is happening here. This was a abstract method and all other implementation were using merge operation on parsed.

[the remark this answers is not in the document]
suresh-prakash: I think that's fine. 
It is to handle multi-value cases.
Basically, if you send a SubDocumentValue.ofNumbers(...), it would add all those numbers to that field.
Let's leave it as is for now.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -2433,6 +2461,71 @@ void testSameHierarchyUpdateThrowsException(final String datastoreName) throws I
...
+          SubDocumentUpdate.builder()
+              .subDocument("props.added.list")
+              .operator(ADD)
+              .subDocumentValue(SubDocumentValue.of(new Integer[] {5, 1, 5}))
+              .build();
+      final Query query_addList = Query.builder().build();
+      final List<SubDocumentUpdate> updates_addList = List.of(addList);
+      assertExceptionForNonNumericValues(collection, datastoreName, query_addList, updates_addList);
+
+      // assert exception for Object
```
suresh-prakash, line 2495: Nice separation of testing different non-numeric values.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -2433,6 +2461,71 @@ void testSameHierarchyUpdateThrowsException(final String datastoreName) throws I
...
+      final Query query_addObject = Query.builder().build();
+      final List<SubDocumentUpdate> updates_addObject = List.of(addObject);
+      assertExceptionForNonNumericValues(
+          collection, datastoreName, query_addObject, updates_addObject);
+    }
+
+    private void assertExceptionForNonNumericValues(
+        Collection collection, String datastoreName, Query query, List<SubDocumentUpdate> updates) {
+      if (MONGO_STORE.equals(datastoreName)) {
+        assertThrows(
```
suresh-prakash, line 2516: Any reason why "IOException" can't be thrown for MongoDB?
The consumers of the library might not know (and hence might not catch) MongoWriteException.

If it's not possible, probably, we need to fix it to throw IOException.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/update/parser/PostgresAddValueParser.java
```diff
@@ -0,0 +1,26 @@
...
+  public String parseLeaf(UpdateParserInput input) {
+    final Params.Builder paramsBuilder = input.getParamsBuilder();
+    final PostgresSubDocumentValueParser valueParser =
+        new PostgresSubDocumentValueParser(paramsBuilder);
+
+    paramsBuilder.addObjectParam(formatSubDocPath(input.getPath()[0]));
+    final String parsedValue = input.getUpdate().getSubDocumentValue().accept(valueParser);
+    return String.format(
+        "jsonb_set(%s, ?::text[], (COALESCE(%s->>'%s', '0')::float + %s::float)::text::jsonb)",
+        input.getBaseField(), input.getBaseField(), input.getPath()[0], parsedValue);
```
suresh-prakash, line 24: There's a utility method for field access (->>).
We could directly use that to avoid duplicating.
```suggestion
    final String fieldAccess = prepareFieldDataAccessorExpr(input.getPath()[0], input.getBaseField());
    return String.format(
        "jsonb_set(%s, ?::text[], (COALESCE(%s, '0')::float + %s::float)::text::jsonb)",
        input.getBaseField(), fieldAccess, parsedValue);
```

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/update/MongoUpdateExecutor.java
```diff
@@ -111,8 +111,14 @@ public Optional<MongoCursor<BasicDBObject>> bulkUpdate(
...
-    log.debug("Updating {} using {} with filter {}", collection.getNamespace(), setObject, filter);
-    collection.updateMany(filter, setObject);
+  private void logAndUpdate(final BasicDBObject filter, final BasicDBObject setObject)
+      throws IOException {
+    try {
+      log.debug(
+          "Updating {} using {} with filter {}", collection.getNamespace(), setObject, filter);
+      collection.updateMany(filter, setObject);
+    } catch (Exception e) {
+      throw new IOException("Error while uploading", e);
```
suresh-prakash, line 121:
```suggestion
      throw new IOException("Error while updating", e);
```

suresh-prakash (approved):

suresh-prakash (approved):

PR #157, opened 2023-06-21
Introduce connection config and builders

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDefaults.java
```diff
@@ -3,11 +3,8 @@
 import java.time.Duration;
 
 public interface PostgresDefaults {
-  String DEFAULT_USER = "postgres";
-  String DEFAULT_PASSWORD = "postgres";
-  String DEFAULT_DB_NAME = "postgres";
 
-  int DEFAULT_MAX_CONNECTION_ATTEMPTS = 200;
+  int DEFAULT_MAX_CONNECTION_ATTEMPTS = 5;
```
aaron-steinfeld, line 7: 200  to 5? why the dramatic change?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDefaults.java
```diff
@@ -3,11 +3,8 @@
 import java.time.Duration;
 
 public interface PostgresDefaults {
-  String DEFAULT_USER = "postgres";
```
aaron-steinfeld, line 6: Not sure if these were refactored or removed - any removal of default behavior would be a breaking change we should avoid.

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/TypesafeConfigConnectionConfigExtractor.java
```diff
@@ -0,0 +1,106 @@
...
+
+  private TypesafeConfigConnectionConfigExtractor(
+      @NonNull final Config config, @NonNull final DatabaseType type) {
+    this.config = config;
+    this.connectionConfigBuilder = ConnectionConfig.builder().type(type);
+    this.connectionCredentialsBuilder = ConnectionCredentials.builder();
+    this.connectionPoolConfigBuilder = ConnectionPoolConfig.builder();
+  }
+
+  public static TypesafeConfigConnectionConfigExtractor from(
```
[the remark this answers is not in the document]
suresh-prakash: Constructors have private access. Only static methods are exposed.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDefaults.java
suresh-prakash: Reverted.

aaron-steinfeld (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/Datastore.java
```diff
@@ -3,11 +3,16 @@
...
 import java.util.Set;
+import org.hypertrace.core.documentstore.model.config.ConnectionConfig;
 
 public interface Datastore {
 
+  /** Deprecated in favour of {@link Datastore#init(ConnectionConfig)} */
+  @Deprecated(forRemoval = true)
   boolean init(Config datastoreConfig);
 
+  void init(final ConnectionConfig connectionConfig);
```
kotharironak, line 14: Will these change the way we provide config in application.conf files wherever this lib is used? Or its a subset of existing config?

document-store/src/main/java/org/hypertrace/core/documentstore/DatastoreProvider.java
```diff
@@ -1,45 +1,56 @@
...
-    Class<? extends Datastore> clazz = registry.get(type.toLowerCase());
-    try {
-      Constructor<? extends Datastore> constructor = clazz.getConstructor();
-      Datastore instance = constructor.newInstance();
-      instance.init(config);
-      return instance;
-    } catch (Exception e) {
-      throw new IllegalArgumentException("Exception creating DocDatastore", e);
-    }
+  @SuppressWarnings("unused")
```
kotharironak, line 36: Main motivation here is to move away from `typesafe` Config?

document-store/src/main/java/org/hypertrace/core/documentstore/Datastore.java
kotharironak: I am assuming that main motivation here that to move away from typesafe config, right?

suresh-prakash: Yes. To decouple typesafe config from document-store model

kotharironak: For now, we have all the config related to connection - `ConnectionConfig`. and also so far we didn't find a need for other config. Bug, was thinking, would `DocStoreConfig` have been more generic that include `connection` + any other config?

Just a thought, I guess, you would have already consider that, right?

suresh-prakash: That makes sense. Wrapped `ConnectionConfig` inside `DatastoreConfig` for datastore-level configurations. (so that we can later introduce `CollectionConfig` also for collection-level configurations).

document-store/src/main/java/org/hypertrace/core/documentstore/DatastoreProvider.java
suresh-prakash: Yup. Make it a first-class configuration, which can be generated from any source.
The type-safe to ConnectionConfig conversion is provided via a Utility class.

kotharironak (approved):
LGTM.

kotharironak:
@suresh-prakash Some integration tests are failing due to change.

document-store/src/main/java/org/hypertrace/core/documentstore/Datastore.java
```diff
@@ -1,13 +1,9 @@
 package org.hypertrace.core.documentstore;
 
-import com.typesafe.config.Config;
 import java.util.Map;
 import java.util.Set;
 
 public interface Datastore {
-
-  boolean init(Config datastoreConfig);
```
kotharironak, line 9: As this was never used except internally via DataStoreProvider, we thought of removing it. 
Its not use anywhere, right?

suresh-prakash: Yeah. Not getting used from outside clients.

kotharironak (approved):

PR #164, opened 2023-07-20
chore: added exists integration tests

saxenakshitiz:
There are two operator: EXISTS and NOT_EXITS
- EXISTS : check if the give field exists
- NOT_EXISTS: check if the give field doesn't exists

Ideally, above two operator doesn't requires the rhs expression.  But, being part of RelationOperator family at docstore level they takes argument as either true or false. Here, client can provide rhs expression either as string or boolean.

In mongo impl, both were taken care by not considering rhs expression.
In postgres impl of older `search` API, it was translated to `IS NULL` or `IS NOT NULL` of JSONB query syntax using field accessor pattern. However, the same was missing in Postgres Impl of `find` or `aggregate` API where the above operators were not added into field accessor group.

This PR fix that.


document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -334,12 +338,192 @@ public void testWithDifferentFieldTypes(String dataStoreName) throws Exception {
...
     }
-    Assertions.assertEquals(2, documents.size());
+    Assertions.assertEquals(3, documents.size());
+
+    datastore.deleteCollection(COLLECTION_NAME);
+  }
+
+  @ParameterizedTest
+  @MethodSource("databaseContextProvider")
+  public void testExistsOperatorWithSearch(String dataStoreName) throws Exception {
```
kotharironak, line 353: How is this different from `testExistsFilter` test case?
https://github.com/hypertrace/document-store/blob/main/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java#L1960

saxenakshitiz: It is same. Will remove it.

saxenakshitiz: removed

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -334,12 +338,209 @@ public void testWithDifferentFieldTypes(String dataStoreName) throws Exception {
...
+            ImmutablePair.of("isCostly", true)));
+
+    org.hypertrace.core.documentstore.query.Query.QueryBuilder queryBuilder;
+    // query exists
+    queryBuilder = org.hypertrace.core.documentstore.query.Query.builder();
+    queryBuilder.setFilter(
+        RelationalExpression.of(
+            IdentifierExpression.of("name"),
+            RelationalOperator.EXISTS,
+            ConstantExpression.of("true")));
```
suresh-prakash, line 398:
```suggestion
            ConstantExpression.of(true)));
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -334,12 +338,209 @@ public void testWithDifferentFieldTypes(String dataStoreName) throws Exception {
...
+      documents.add(results.next());
+    }
+    Assertions.assertEquals(4, documents.size());
+
+    queryBuilder = org.hypertrace.core.documentstore.query.Query.builder();
+    queryBuilder.setFilter(
+        RelationalExpression.of(
+            IdentifierExpression.of("junkKey"),
+            RelationalOperator.EXISTS,
+            ConstantExpression.of("true")));
```
suresh-prakash, line 411: Nit:
```suggestion
            ConstantExpression.of(true)));
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -334,12 +338,209 @@ public void testWithDifferentFieldTypes(String dataStoreName) throws Exception {
...
+    }
+    Assertions.assertEquals(0, documents.size());
+
+    // queryBuilder not exists
+    queryBuilder = org.hypertrace.core.documentstore.query.Query.builder();
+    queryBuilder.setFilter(
+        RelationalExpression.of(
+            IdentifierExpression.of("name"),
+            RelationalOperator.NOT_EXISTS,
+            ConstantExpression.of("true")));
```
suresh-prakash, line 425:
```suggestion
            ConstantExpression.of(true)));
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -334,12 +338,209 @@ public void testWithDifferentFieldTypes(String dataStoreName) throws Exception {
...
+      documents.add(results.next());
+    }
+    Assertions.assertEquals(0, documents.size());
+
+    queryBuilder = org.hypertrace.core.documentstore.query.Query.builder();
+    queryBuilder.setFilter(
+        RelationalExpression.of(
+            IdentifierExpression.of("junkKey"),
+            RelationalOperator.NOT_EXISTS,
+            ConstantExpression.of("true")));
```
suresh-prakash, line 438:
```suggestion
            ConstantExpression.of(true)));
```

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: There are two different sets, one takes string as input and other boolean expression.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: There are two different sets, one takes string as input and other boolean expression.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: There are two different sets, one takes string as input and other boolean expression.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
kotharironak: There are two different sets, one takes string as input and other boolean expression.

suresh-prakash (approved):

PR #166, opened 2023-08-04
Added support for multiple endpoints and replica-set

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/model/config/ConnectionConfig.java
```diff
@@ -38,11 +40,11 @@ public static ConnectionConfigBuilder builder() {
   @FieldDefaults(level = AccessLevel.PRIVATE)
   public static class ConnectionConfigBuilder {
     DatabaseType type;
-    String host = DEFAULT_HOST;
-    Integer port;
+    List<Endpoint> endpoints = new ArrayList<>();
     String database;
     ConnectionCredentials credentials;
     String applicationName = DEFAULT_APP_NAME;
+    String replicaSet;
```
kotharironak, line 47: - How was this pass in earlier version where we used to pass only TypeSafe Config?
- Was it not required prior?

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/ConnectionConfig.java
```diff
@@ -38,11 +40,11 @@ public static ConnectionConfigBuilder builder() {
   @FieldDefaults(level = AccessLevel.PRIVATE)
   public static class ConnectionConfigBuilder {
     DatabaseType type;
-    String host = DEFAULT_HOST;
-    Integer port;
+    List<Endpoint> endpoints = new ArrayList<>();
```
kotharironak, line 43: Did we update any service that uses it directly (the new API)? if, yes, this change will cause compiler error on that service on upgrade of library, right?

document-store/src/main/java/org/hypertrace/core/documentstore/TypesafeDatastoreConfigAdapter.java
```diff
@@ -31,13 +34,32 @@ public DatastoreConfig convert(final Config config) {
...
-              connectionConfig.port(),
+              connectionConfig.endpoints(),
               connectionConfig.database(),
               connectionConfig.credentials(),
-              connectionConfig.applicationName()) {
-            @Override
-            public ConnectionString toConnectionString() {
+              connectionConfig.applicationName(),
+              null) {
+            public MongoClientSettings toSettings() {
```
kotharironak, line 42: What is change for `toSettings` ? Def looks same what is there in class `MongoConnectionConfig`.

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/TypesafeConfigDatastoreConfigExtractor.java
```diff
@@ -49,16 +53,41 @@ public static TypesafeConfigDatastoreConfigExtractor from(
...
+    }
+
+    for (final Config endpointConfig : config.getConfigList(endpointsBaseKey)) {
+      final Endpoint.EndpointBuilder builder = Endpoint.builder();
+
+      if (endpointConfig.hasPath(hostKey)) {
+        builder.host(endpointConfig.getString(hostKey));
+      }
+
+      if (endpointConfig.hasPath(portKey)) {
```
kotharironak, line 71: If its missing, will it use default port?

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/ConnectionConfig.java
suresh-prakash: Yeah. Not required earlier.

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/ConnectionConfig.java
suresh-prakash: > Did we update any service that uses it directly (the new API)? 

No.

> if, yes, this change will cause compiler error on that service on upgrade of library, right?

Even if so, it would not since the public methods haven't changed. Only the internal representation and the way it is consumed in the library has changed.

document-store/src/main/java/org/hypertrace/core/documentstore/TypesafeDatastoreConfigAdapter.java
suresh-prakash: This is overriding the implementation in `MongoConnectionConfig`.
The original implementation is consuming the configs directly. And, the overridden implementation is generating a connection string and using it.

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/TypesafeConfigDatastoreConfigExtractor.java
suresh-prakash: Yes. Mongo and Postgres constructors make sure to fill-in the respective default ports if missing.

kotharironak (approved):

PR #167, opened 2023-08-04
Introduce metric exporters

suresh-prakash:
Metric store support from the library:
- A new metric provider interface to get the database-type-specific metrics and common custom metrics
- For each supported database (MongoDB and Postgres), the number of connections for the connecting application name is implemented
- A common implementation includes returning of the custom metrics upon invocation


kotharironak:
What DB level metrics and logging are available?

document-store/src/main/java/org/hypertrace/core/documentstore/metric/mongo/MongoDocStoreMetricProvider.java
```diff
@@ -0,0 +1,68 @@
...
+          DocStoreMetric.builder()
+              .name(NUM_ACTIVE_CONNECTIONS_METRIC_NAME)
+              .value(count)
+              .labels(Map.of(APP_NAME_LABEL, applicationNameInCurrentConnection))
+              .build();
+      log.debug("Returned metric: {}", metric);
+      return metric;
+    } catch (final Exception e) {
+      log.error("Unable to capture {}", NUM_ACTIVE_CONNECTIONS_METRIC_NAME, e);
+      return DocStoreMetric.builder()
```
kotharironak, line 62: This means in any error case, we are saying, there are `0` connections. But, we actually don't know are there no connection or there are `x` connection. How is miss handled in metrics world - in our case in Prometheus?

E.g with this impl, lets say client reports metrics every 1 min:
first call: x connection -> prometheus: x connection total_count = x
second call: 0 connection due to error -> prometheus: 0 connection total_count = x
third call: x connection -> prometheus : x connection total_count = 2x

Or, if we return `null`, and client doesn't report.
first call: x connection -> prometheus: x connection total_count = x
second call: null due to error  and not reported to prometheus vai client: prom exporter will still read the value x connection : total_count = x
third call : x connection -> prometheus x connection total_count = 2x

But, here there will be a flat line. And above case, it will be deep to zero. I think, both cases are handled in Prometheus for counter metrics.

So, I think, we should be fine here.

document-store/src/main/java/org/hypertrace/core/documentstore/metric/postgres/PostgresDocStoreMetricProvider.java
```diff
@@ -0,0 +1,82 @@
...
+        final DocStoreMetric metric =
+            metrics.get(0).toBuilder()
+                .labels(Map.of(APP_NAME_LABEL, applicationNameInCurrentConnection))
+                .build();
+        log.debug("Returned metric: {}", metric);
+        return metric;
+
+      default:
+        log.error(
+            "Found {} values for metric {}", metrics.size(), NUM_ACTIVE_CONNECTIONS_METRIC_NAME);
```
kotharironak, line 78: Like above, its error scenario. Shall we include `Could not report metric {} as Found multiple values {}`?

document-store/src/main/java/org/hypertrace/core/documentstore/metric/postgres/PostgresDocStoreMetricProvider.java
```diff
@@ -0,0 +1,82 @@
...
+import org.hypertrace.core.documentstore.query.Query;
+import org.hypertrace.core.documentstore.query.SelectionSpec;
+
+@Slf4j
+public class PostgresDocStoreMetricProvider extends BaseDocStoreMetricProviderImpl {
+  private static final String NUM_ACTIVE_CONNECTIONS_METRIC_NAME =
+      "num.active.postgres.connections";
+  private static final String APP_NAME_LABEL = "app_name";
+  private static final String APPLICATION_COLUMN_NAME = "application_name";
+  private static final String PG_STAT_ACTIVITY_TABLE = "pg_stat_activity";
```
kotharironak, line 28: Do we need to access this as `postgres` user?

document-store/src/main/java/org/hypertrace/core/documentstore/metric/BaseDocStoreMetricProviderImpl.java
```diff
@@ -0,0 +1,93 @@
...
+              customMetricConfig.metricName(),
+              query);
+          continue;
+        }
+
+        final long metricValue;
+        if (node.has(VALUE_KEY)) {
+          metricValue = Double.valueOf(node.get(VALUE_KEY).doubleValue()).longValue();
+        } else {
+          log.warn(
```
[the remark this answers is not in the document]
suresh-prakash: > Is this actionable in some way? If, not, should we change to debug?

When it hits this code-path, it means the input query is built incorrectly. Hence, the `warn` level logging.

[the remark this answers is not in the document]
suresh-prakash: > Is this possible scenario that few returned documents are missing with metric_value?

Not possible as it should be a selection.

> The aggregate query will be metric query for this api, right? Something like:
select count(*) as metric_value from document or
select count(*) as metric_value from document group by document.foo.bar like "XYZ"

Yup. If the query is selected as `select count(*) as metric_value, document.foo.bar AS bar from document group by document.foo.bar`, each unique value of bar becomes a separate metric with a different label.

document-store/src/main/java/org/hypertrace/core/documentstore/metric/DocStoreMetric.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Value;
+import lombok.experimental.Accessors;
+import org.checkerframework.checker.index.qual.NonNegative;
+
+@Value
+@Builder(toBuilder = true)
+@Accessors(fluent = true)
+public class DocStoreMetric {
+  @NonNull String name;
+  @Default @NonNegative long value = 0;
```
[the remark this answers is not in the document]
suresh-prakash: Oh! Yeah. We are registering them as `gauge` and metrics like `AVG(col)` may contain double values. I'll update it to double.

document-store/src/main/java/org/hypertrace/core/documentstore/metric/postgres/PostgresDocStoreMetricProvider.java
suresh-prakash: Will be able to as long as necessary permissions are present. Just want to avoid creating a different connection for this purpose.

PR #169, opened 2023-09-11
Introduce the atomic create or replace API to overcome the limitations of upsert

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -249,6 +293,25 @@ public Document upsertAndReturn(Key key, Document document) throws IOException {
     return dbObjectToDocument(upsertResult);
   }
 
+  @SuppressWarnings("MismatchedQueryAndUpdateOfCollection")
+  private List<BasicDBObject> prepareForCreateOrReplace(final Key key, final Document document)
+      throws JsonProcessingException {
+    final long now = System.currentTimeMillis();
+
+    // Explicit createdTime field projection automatically unsets all the other existing fields
```
kotharironak, line 301: Which other fields? Referring only to CREATED_TIME?

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -30,8 +30,11 @@ public interface Collection {
    *
    * @param key Unique key of the document in the collection.
    * @param document Document to be upserted.
-   * @return True if this operation resulted in update of an existing document. False, otherwise.
+   * @return True if this operation resulted in an update of an existing document or creation of a
+   *     new document.
+   * @deprecated Use {@link #createOrReplace} for a consistent behavior.
    */
+  @Deprecated(forRemoval = true)
```
kotharironak, line 37: What is the difference between `upsert` and `createOrReplace`?
Are we facing some issue with the existing api? if, yes, what?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
suresh-prakash: All the other fields in the document are unset.
For example, if the existing document was
```json
{
   "foo": "bar",
   "createdTime": 1325374949,
   "updatedTime": 1325374949
}
```
projecting `createdTime` will automatically unset `foo` and `updatedTime`.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
suresh-prakash: Yes. The issue is mentioned in the documentation comments of the `upsert` API.
For example, upserting 
```json
{
    "foo2": "bar2"
}
```
into an existing document
```json
{
    "foo1": "bar1"
}
```
results in 
```json
{
    "foo2": "bar2"
}
```
correctly in Postgres, but, results in
```json
{
    "foo1": "bar1",
    "foo2": "bar2"
}
```
in Mongo implementation.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -249,6 +293,25 @@ public Document upsertAndReturn(Key key, Document document) throws IOException {
...
+    final BasicDBObject project =
+        new BasicDBObject(
+            PROJECT_CLAUSE,
+            new BasicDBObject(
+                CREATED_TIME,
```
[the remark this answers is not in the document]
suresh-prakash: Well. The `upsert` uses the `replacement` document version of the `upsertOne` or `findOneAndReplace`. The `createOrReplace` uses the `aggregation pipeline` version of the `upsertOne` or `findOneAndReplace`.
To overcome the issue mentioned @ https://github.com/hypertrace/document-store/pull/169#discussion_r1323102021

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
kotharironak: - Was `updateSubDoc` capable to solve this problem? 
- I guess `upsert` was meant for full document update. Is the requirement to add some additional stuff in the existing document?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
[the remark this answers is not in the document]
kotharironak: Got it.

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
suresh-prakash: > Was updateSubDoc capable to solve this problem?

Nopes. There also, I need to explicitly unset the other top-level fields that are undesired.

> I guess upsert was meant for full document update.

Yes. It was supposed to completely replace the existing document. But, that was not happening in the case of MongoDB.

> Is the requirement to add some additional stuff in the existing document?

Nope. The new API targets at fixing the bug in the Mongo implementation of upsert.

kotharironak: As discuss offline, we need a way to move to usage of newer APIs. So, one of the way to define collectionV2.

kotharironak (approved):

PR #170, opened 2023-09-19
Safe handle non-numeric port by skipping the consumption

suresh-prakash:


aaron-steinfeld (dismissed):

document-store/src/main/java/org/hypertrace/core/documentstore/model/config/TypesafeConfigDatastoreConfigExtractor.java
```diff
@@ -78,80 +87,80 @@ public TypesafeConfigDatastoreConfigExtractor keysForEndpoints(
...
-    if (config.hasPath(key)) {
+    if (config.hasPath(key) && STRING.equals(config.getValue(key).valueType())) {
       endpointBuilder.host(config.getString(key));
     }
     return this;
   }
 
   public TypesafeConfigDatastoreConfigExtractor portKey(@NonNull final String key) {
-    if (config.hasPath(key)) {
+    if (config.hasPath(key) && NUMBER.equals(config.getValue(key).valueType())) {
```
kotharironak, line 97: What would happen if the port value is `STRING`? would that is not parsed?
Most of the places for config, we already know the types as config is driven manually.

suresh-prakash: If the port is a string, it throws an exception without an explicit type check. Hence, added the safety type-check everywhere in addition to the existence check.

kotharironak: So far, we didn't have type checking in the code, so if type checked is enforced, it should throw the exception if desired type is not matched.

And if there are multiple way to provides same configuration, then we should throw an exception if none of the valid configuration found.  E.g if say for connection, it can be configured either using
- connectionString
- host + port
So, if both are invalids, we should throw the exception. It's better to exit early with an error rather than discovering the issue while executing the query that connection error, etc.

suresh-prakash: Makes sense. Updated the PR to reactively consume "host" and "port" for older configuration maintaining backward compatibility and reverted the change for type-checking.

document-store/src/main/java/org/hypertrace/core/documentstore/TypesafeDatastoreConfigAdapter.java
```diff
@@ -19,43 +19,14 @@ interface TypesafeDatastoreConfigAdapter {
...
-                  .connectionConfig();
-
       final MongoConnectionConfig overridingConnectionConfig =
-          new MongoConnectionConfig(
-              connectionConfig.endpoints(),
-              connectionConfig.database(),
-              connectionConfig.credentials(),
-              connectionConfig.applicationName(),
-              null) {
+          new MongoConnectionConfig(emptyList(), null, null, "", null) {
```
kotharironak, line 23: So, now it will work with old config if connectionUrl is provided, right?

suresh-prakash: Yes.

kotharironak (approved):

PR #173, opened 2023-09-26
Array filtering support (Part #1): API and Mongo filter refactoring

suresh-prakash:
So far, we do not have support for filtering by LHS arrays.
The PRs in this series aim to support LHS arrays in filters.

An LHS array can contain either primitives or documents.
In the case of a primitive LHS array, we can apply a relational operator on each element.
In the case of a document LHS array, we can apply a generic filter (including relational expression or logical expression or even another array filter expression)

Sample data: 
```json
[
  {
    "key": 3,
    "planets": [
      {
        "name": "Planet 1",
        "elements": [
          "Oxygen",
          "Water",
          "Nitrogen"
        ]
      },
      {
        "name": "Planet 2",
        "elements": [
          "Oxygen",
          "Helium",
          "Water"
        ]
      }
    ]
  },
  {
    "key": 1,
    "planets": [
      {
        "name": "Mercury"
      },
      {
        "name": "Venus"
      },
      {
        "name": "Earth",
        "elements": [
          "Oxygen",
          "Nitrogen"
        ]
      },
      {
        "name": "Mars",
        "elements": [
          "Iron"
        ]
      },
      {
        "name": "Jupiter",
        "elements": []
      },
      {
        "name": "Saturn",
        "elements": []
      },
      {
        "name": "Uranus",
        "elements": [
          "Hydrogen",
          "Helium",
          "Methane"
        ]
      },
      {
        "name": "Neptune",
        "elements": [
          "Hydrogen",
          "Helium",
          "Methane",
          "Ammonia"
        ]
      }
    ]
  },
  {
    "key": 2,
    "planets": [
      {
        "name": "Mercury",
        "elements": [
          "Silicate",
          "Aluminum"
        ]
      },
      {
        "name": "Venus",
        "elements": [
          "Carbon Dioxide",
          "Sulfuric Acid"
        ]
      },
      {
        "name": "Earth",
        "elements": [
          "Oxygen",
          "Nitrogen",
          "Water"
        ]
      },
      {
        "name": "Mars",
        "elements": [
          "Iron",
          "Silicate"
        ]
      },
      {
        "name": "Jupiter",
        "elements": [
          "Hydrogen",
          "Helium"
        ]
      },
      {
        "name": "Saturn",
        "elements": [
          "Hydrogen",
          "Helium",
          "Methane"
        ]
      },
      {
        "name": "Uranus",
        "elements": [
          "Hydrogen",
          "Helium",
          "Methane",
          "Ammonia"
        ]
      },
      {
        "name": "Neptune",
        "elements": [
          "Hydrogen",
          "Helium",
          "Methane",
          "Ammonia",
          "Methane Hydrate"
        ]
      }
    ]
  }
]
```

Input query: `Get a list of solar systems with at least one planet not having either oxygen or water.`
```sql
SELECT * 
FROM solar_systems 
WHERE ANY(planets) [NOT ((ANY(elements) = "Oxygen") OR (ANY(elements) = "Water"))]
```

The filter API would look like
```java
FilterTypeExpression filter = DocumentArrayFilterExpression.builder()
    .operator(ANY)
    .arraySource(IdentifierExpression.of("planets"))
    .filter(
       LogicalExpression.builder()
           .operator(NOT)
           .operand(
                 LogicalExpression.builder()
                       .operator(OR)
                       .operand(
                             ArrayRelationalFilterExpression.builder()
                                  .operator(ANY)
                                  .filter(
                                       RelationalExpression.of(
                                            IdentifierExpression.of("elements"),
                                            EQ,
                                            ConstantExpression.of("Oxygen"))))
                        .operand(
                             ArrayRelationalFilterExpression.builder()
                                  .operator(ANY)
                                  .filter(
                                       RelationalExpression.of(
                                            IdentifierExpression.of("elements"),
                                            EQ,
                                            ConstantExpression.of("Water"))))))
    .build();
```

Sample MongoDB match stage query to be generated
```json
{
  "$expr":
  {
    "$anyElementTrue":
    {
      "$map":
      {
        "input":
        {
          "$ifNull": [
            "$planets",
            []
          ]
        },
        "as": "planet",
        "in":
        {
          "$not":
          {
            "$or": [
              {
                "$anyElementTrue":
                {
                  "$map":
                  {
                    "input":
                    {
                      "$ifNull": [
                        "$$planet.elements",
                        []
                      ]
                    },
                    "as": "element",
                    "in":
                    {
                      "$eq": ["$$element","Oxygen"]
                    }
                  }
                }
              },
              {
                "$anyElementTrue":
                {
                  "$map":
                  {
                    "input":
                    {
                      "$ifNull": [
                        "$$planet.elements",
                        []
                      ]
                    },
                    "as": "element",
                    "in":
                    {
                      "$eq": ["$$element","Water"]
                    }
                  }
                }
              }
            ]
          }
        }
      }
    }
  }
}
```

Sample Postgres Query:
```sql
SELECT *  
FROM solar_systems  
WHERE 
EXISTS 
(SELECT 1 
 FROM  jsonb_array_elements(COALESCE(str->'planets', '[]'::jsonb)) AS planet 
 WHERE 
 NOT 
 (
 EXISTS 
 (SELECT 1 
  FROM jsonb_array_elements(COALESCE(planet->'elements', '[]'::jsonb)) AS elements 
  WHERE TRIM('"' FROM elements::text) = 'Oxygen'
 )
 OR 
 EXISTS 
 (SELECT 1 
  FROM jsonb_array_elements(COALESCE(planet->'elements', '[]'::jsonb)) AS elements 
  WHERE TRIM('"' FROM elements::text) = 'Water'
 )
 )
);
```

**Notes:**

1. Only `ANY` operator is supported for now
2. A future PR would introduce a `NOT` logical operator
3. `ALL` can be achieved by using `NOT` twice to the `ANY` filter (before and after). E.g.: `Fetch solar systems with all planets containing both oxygen and water` is equivalent to `Fetch solar systems with no planet without both oxygen and water`
4. Similarly `NONE` can be achieved by using `NOT` once before `ANY`, because `NONE = NOT ANY`
5. The examples given in this description are for illustrative purposes only. For practical applications, the queries would be simpler (E.g.: Use `IN` instead of `OR` and `=`, static import `not` and `or` logical operators, etc.)

**This PR contains**
* API changes to support array filtering
* Refactor Mongo relational filter parser


SrikarMannepalli (approved):

kotharironak:
@suresh-prakash 

We already had support for unnest expressions to flatten array columns. Does the new expression support provides any missing feature? If not, was the change made to address any performance issues?

I think the example query, such as 'solar_system,' could have been expressed using unnest, correct?

suresh-prakash:
> @suresh-prakash
> 
> We already had support for unnest expressions to flatten array columns. Does the new expression support provides any missing feature? If not, was the change made to address any performance issues?
> 
> I think the example query, such as 'solar_system,' could have been expressed using unnest, correct?

The Unnest expression already splits the document and then applies the filter. So, the filtered document is only partial (as it doesn't contain the array elements that didn't match the unnest filter). 
This array filter support is to outcome that issue to filter the documents without splitting them.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/ArrayRelationalFilterExpression.java
```diff
@@ -0,0 +1,56 @@
...
+ *        IN,
+ *        ConstantExpression.ofStrings("Blue", "Green")
+ *    )
+ *    .build();
+ * </code>
+ */
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class ArrayRelationalFilterExpression implements FilterTypeExpression {
```
kotharironak, line 32: Shouldn't the DocumentArrayFilterExpression suffice? Also DocumentArrayFilterExpression explicitly mark the array source. Here, we will have to deduce (or implicit assumption).

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/DocumentArrayFilterExpression.java
```diff
@@ -0,0 +1,68 @@
...
+ *      )
+ *    .build();
+ */
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DocumentArrayFilterExpression implements FilterTypeExpression {
+  ArrayOperator operator;
+
+  SelectTypeExpression arraySource;
```
kotharironak, line 45: Q: Could this be anything else other than `IdentifierExpression`?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/DocumentArrayFilterExpression.java
```diff
@@ -0,0 +1,68 @@
...
+ *          IdentifierExpression.of("name"),
+ *          EQ,
+ *          ConstantExpression.of("Comb"))
+ *      )
+ *    .build();
+ */
+@Value
+@Builder
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DocumentArrayFilterExpression implements FilterTypeExpression {
```
kotharironak, line 42: Other filter expressions are not prefixed with `Document`, any specific reason?

Could this be just called - `ArrayExpression` or `ArrayFilterExpression`?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/ArrayRelationalFilterExpression.java
suresh-prakash: > Shouldn't the DocumentArrayFilterExpression suffice?

We would want to treat primitive arrays and document (object) arrays differently, because
1. In the case of primitive arrays, we do not have the LHS and have a special parsing mechanism
2. In the case of document arrays, we always need the LHS and its path is relative to the array source

> Also DocumentArrayFilterExpression explicitly mark the array source. Here, we will have to deduce (or implicit assumption).

Yes. That's by design. For primitive arrays, the LHS itself is the array source (so we don't need to explicitly ask the same from the users again).

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFilterTypeExpressionParser.java
```diff
@@ -7,27 +7,41 @@
...
   public Map<String, Object> visit(final LogicalExpression expression) {
-    return new MongoLogicalExpressionParser().parse(expression);
+    return new MongoLogicalExpressionParser(relationalFilterContext).parse(expression);
   }
 
   @SuppressWarnings("unchecked")
   @Override
   public Map<String, Object> visit(final RelationalExpression expression) {
-    return new MongoRelationalExpressionParser().parse(expression);
+    return new MongoRelationalExpressionParser().parse(expression, relationalFilterContext);
```
kotharironak, line 44: Any specific reason to parse `context` as part of method `parse`. But on the above, the context is passed as part of constructor - `MongoLogicalExpressionParser`. If, the object is supposed to run always in the context, we can pass it in constructor.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/DocumentArrayFilterExpression.java
suresh-prakash: For now, we only support the `IdentifierExpression`. Anything other than that will throw an exception from the `document-store` before going to the database. I deliberately made it `SelectTypeExpression` to support extensibility (E.g.: We want to apply filtering on some function that generates an array)

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/DocumentArrayFilterExpression.java
suresh-prakash: > any specific reason?

Yeah. To clearly distinguish between a primitive-array and a document-array (object-array).
Calling it a document array instead of an object array, because it supports filtering by nested fields too.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFilterTypeExpressionParser.java
suresh-prakash: Makes sense. Sure, it can be passed in the constructor. I'll update it.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/filter/MongoContainsRelationalFilterParser.java
```diff
@@ -0,0 +1,26 @@
+package org.hypertrace.core.documentstore.mongo.query.parser.filter;
+
+import com.mongodb.BasicDBObject;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import org.hypertrace.core.documentstore.expression.impl.RelationalExpression;
+import org.hypertrace.core.documentstore.mongo.query.parser.filter.MongoRelationalFilterParserFactory.MongoRelationalFilterContext;
+
+@AllArgsConstructor
+public class MongoContainsRelationalFilterParser implements MongoRelationalFilterParser {
```
kotharironak, line 10: `MongoContainsRelationalFilterParser` -> `MongoEqualRelationalFilterParser` or `MongoElementMatchRelationalFilterParser` ?

kotharironak: Ignore this one, next I saw that we have enum for `CONTAINS` and so.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/ArrayRelationalFilterExpression.java
kotharironak: For primitive types: would name `ArrayFilterExpression` suffice?
For non-primitive : `DocumentArrayFilterExpression`?

kotharironak: So, I am assuming that naming is to do with that we just want to allow only relation expression in primitive array types.

suresh-prakash: > So, I am assuming that naming is to do with that we just want to allow only relation expression in primitive array types.

Yes, exactly. `DocumentArrayFilterExpression` is an `Expression` to `Filter` `Arrays` containing `Document`s. So, the `Filter` in this context means any filter (could be relational, logical or even another array filter).

`ArrayRelationalFilterExpression` is an `Expression` to apply `RelationalFilter` on `Arrays`. In this context, it is evident that only relational filtering can be applied using this. Deliberately left out the term `Primitive` here, because one could also do an equality comparison of 2D arrays using this filter. (Something like `ANY(2D_array) = Document.of("['hello', 'world']")`)

kotharironak (approved):

PR #174, opened 2023-09-27
Increase the connection surrender timeout to 5 minutes

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoDatastore.java
```diff
@@ -91,6 +93,15 @@ public DocStoreMetricProvider getDocStoreMetricProvider() {
     return docStoreMetricProvider;
   }
 
+  @Override
+  public void close() {
+    try {
+      client.close();
```
kotharironak, line 99: can client be null here by any chance?

document-store/src/main/java/org/hypertrace/core/documentstore/Datastore.java
```diff
@@ -17,4 +17,6 @@ public interface Datastore {
 
   @SuppressWarnings("unused")
   DocStoreMetricProvider getDocStoreMetricProvider();
+
+  void close();
```
kotharironak, line 21: Is this interface not exposed outside, right?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
```diff
@@ -51,6 +51,23 @@ public Connection getPooledConnection() throws SQLException {
     return connectionPool.getConnection();
   }
 
+  public void close() {
+    if (connection != null) {
+      try {
+        log.info("Closing connection #{} to {}", count, connectionConfig.toConnectionString());
```
kotharironak, line 57: What the `count` variable here? number of connection in the pool?

kotharironak, line 57: Do we need this info message?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoDatastore.java
suresh-prakash: Nope. The client is initialized in the constructor and hence, cannot be null.

document-store/src/main/java/org/hypertrace/core/documentstore/Datastore.java
suresh-prakash: Yes. Intended to expose this so that the clients can proactively close the connections at the service stop (if they need to) instead of relying on the DB to terminate the connections.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresClient.java
suresh-prakash: The existing method moved here since it was made public. But the count refers to the # of connections already made.

suresh-prakash: The existing method moved here since it was made public. Maybe, worth considering making it as `debug`?

kotharironak (approved):
lgtm

PR #178, opened 2023-10-21
fix: fix for update and bulk array update operation  in postgres

saxenakshitiz:


suresh-prakash (commented):
Looks good to me.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -570,5 +570,8 @@ public static void enrichPreparedStatementWithParams(
                 log.error("SQLException setting Param. key: {}, value: {}", k, v);
               }
             });
+    if (log.isDebugEnabled()) {
+      log.debug("Executing statement - preparedStatement: {}", preparedStatement);
```
kotharironak, line 574: `preparedStatement` has toString, right?

kotharironak (approved):
lgtm

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
saxenakshitiz: Already checked this part. It works fine.

PR #181, opened 2023-11-01
fix: update support for postgres collection names with schema identif…

aaron-steinfeld:
Add postgres support for a table named like "x.y.z" - which is valid in mongo, but not postgres. This change converts the first segment to be the schema name, and quotes everything after (uniformly, whether needed or not)

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
@@ -98,16 +98,21 @@ public class PostgresCollection implements Collection {
       CloseableIterator.emptyIterator();
 
   private final PostgresClient client;
-  private final String collectionName;
+  private final PostgresTableIdentifier tableIdentifier;
   private final PostgresSubDocumentUpdater subDocUpdater;
   private final PostgresQueryExecutor queryExecutor;
   private final UpdateValidator updateValidator;
 
   public PostgresCollection(final PostgresClient client, final String collectionName) {
+    this(client, PostgresTableIdentifier.parse(collectionName));
+  }
+
+  PostgresCollection(final PostgresClient client, final PostgresTableIdentifier tableIdentifier) {
     this.client = client;
-    this.collectionName = collectionName;
-    this.subDocUpdater = new PostgresSubDocumentUpdater(new PostgresQueryBuilder(collectionName));
-    this.queryExecutor = new PostgresQueryExecutor(collectionName);
+    this.tableIdentifier = tableIdentifier;
+    this.subDocUpdater =
+        new PostgresSubDocumentUpdater(new PostgresQueryBuilder(this.tableIdentifier));
+    this.queryExecutor = new PostgresQueryExecutor(this.tableIdentifier);
     this.updateValidator = new CommonUpdateValidator();
   }
 
@@ -369,7 +374,7 @@ public class PostgresCollection implements Collection {
     String updateSubDocSQL =
         String.format(
             "UPDATE %s SET %s=jsonb_set(%s, ?::text[], ?::jsonb) WHERE %s = ?",
-            collectionName, DOCUMENT, DOCUMENT, ID);
+            tableIdentifier, DOCUMENT, DOCUMENT, ID);
     try {
       PreparedStatement preparedStatement =
           client.getConnection().prepareStatement(updateSubDocSQL);
@@ -429,7 +434,7 @@ public class PostgresCollection implements Collection {
   private CloseableIterator<Document> search(Query query, boolean removeDocumentId) {
     String selection = PostgresQueryParser.parseSelections(query.getSelections());
     StringBuilder sqlBuilder =
-        new StringBuilder(String.format("SELECT %s FROM ", selection)).append(collectionName);
+        new StringBuilder(String.format("SELECT %s FROM ", selection)).append(tableIdentifier);
 
     String filters = null;
     Params.Builder paramsBuilder = Params.newBuilder();
@@ -501,7 +506,7 @@ public class PostgresCollection implements Collection {
     try (final Connection connection = client.getPooledConnection()) {
       org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser parser =
           new org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser(
-              collectionName, query);
+              tableIdentifier, query);
       final String selectQuery = parser.buildSelectQueryForUpdate();
 
       try (final PreparedStatement preparedStatement =
@@ -604,7 +609,7 @@ public class PostgresCollection implements Collection {
   public long count(org.hypertrace.core.documentstore.query.Query query) {
     org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser queryParser =
         new org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser(
-            collectionName, query);
+            tableIdentifier, query);
     String subQuery = queryParser.parse();
     String sqlQuery = String.format("SELECT COUNT(*) FROM (%s) p(count)", subQuery);
     try {
@@ -623,7 +628,7 @@ public class PostgresCollection implements Collection {
 
   @Override
   public boolean delete(Key key) {
-    String deleteSQL = String.format("DELETE FROM %s WHERE %s = ?", collectionName, ID);
+    String deleteSQL = String.format("DELETE FROM %s WHERE %s = ?", tableIdentifier, ID);
     try (PreparedStatement preparedStatement = client.getConnection().prepareStatement(deleteSQL)) {
       preparedStatement.setString(1, key.toString());
       preparedStatement.executeUpdate();
@@ -639,10 +644,10 @@ public class PostgresCollection implements Collection {
     if (filter == null) {
       throw new UnsupportedOperationException("Filter must be provided");
     }
-    StringBuilder sqlBuilder = new StringBuilder("DELETE FROM ").append(collectionName);
+    StringBuilder sqlBuilder = new StringBuilder("DELETE FROM ").append(tableIdentifier);
     Params.Builder paramsBuilder = Params.newBuilder();
     String filters = PostgresQueryParser.parseFilter(filter, paramsBuilder);
-    LOGGER.debug("Sending query to PostgresSQL: {} : {}", collectionName, filters);
+    LOGGER.debug("Sending query to PostgresSQL: {} : {}", tableIdentifier, filters);
     if (filters == null) {
       throw new UnsupportedOperationException("Parsed filter is invalid");
     }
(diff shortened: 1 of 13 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -72,7 +73,12 @@ public Set<String> listCollections() {
       DatabaseMetaData metaData = client.getConnection().getMetaData();
       ResultSet tables = metaData.getTables(null, null, "%", new String[] {"TABLE"});
       while (tables.next()) {
-        collections.add(database + "." + tables.getString("TABLE_NAME"));
+        Optional<String> nonPublicSchema =
+            Optional.ofNullable(tables.getString("TABLE_SCHEM"))
```
suresh-prakash, line 77: Typo?
TABLE_SCHEMA

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresTableIdentifier.java
```diff
@@ -0,0 +1,43 @@
+package org.hypertrace.core.documentstore.postgres;
+
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@EqualsAndHashCode
+public class PostgresTableIdentifier {
```
suresh-prakash, line 10: Isn't database a part of the table identifier? Or, it can be skipped because it's already a part of the connection configuration?

aaron-steinfeld: Technically yes. Currently everything is in the scope of one database though, so we omit it. The schema is what we were using the implicit value for ("public").

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
aaron-steinfeld: It's not, but thanks for pointing it out - certain deserves a comment.

> TABLE_CAT String => table catalog (may be null)
TABLE_SCHEM String => table schema (may be null)
TABLE_NAME String => table name
TABLE_TYPE String => table type. Typical types are "TABLE", "VIEW", "SYSTEM TABLE", "GLOBAL TEMPORARY", "LOCAL TEMPORARY", "ALIAS", "SYNONYM".
REMARKS String => explanatory comment on the table
TYPE_CAT String => the types catalog (may be null)
TYPE_SCHEM String => the types schema (may be null)
TYPE_NAME String => type name (may be null)
SELF_REFERENCING_COL_NAME String => name of the designated "identifier" column of a typed table (may be null)
REF_GENERATION String => specifies how values in SELF_REFERENCING_COL_NAME are created. Values are "SYSTEM", "USER", "DERIVED". (may be null)

https://javadoc.io/doc/org.postgresql/postgresql/42.0.0.jre7/org/postgresql/jdbc/PgDatabaseMetaData.html#getTables(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String%5B%5D)

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresTableIdentifier.java
suresh-prakash: Makes sense. Worth mentioning that in the comment too.?

suresh-prakash (approved):
Looks good to me

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -82,6 +95,8 @@ public Set<String> listCollections() {
 
   @Override
   public boolean createCollection(String collectionName, Map<String, String> options) {
+    final PostgresTableIdentifier identifier = PostgresTableIdentifier.parse(collectionName);
+    identifier.getSchema().ifPresent(this::createSchemaIfNotExists);
```
kotharironak, line 99: - If the collectionName is 'x.y', are we considering 'x' as the schema and 'y' as the table name?"

- Would there be a case of the collection name being 'x.y.z'? In that case, the first part would be the schema, and 'y.z' would be the tableName?

So far, we were considering all the tables under `public` schema.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
```diff
@@ -90,32 +105,45 @@ public boolean createCollection(String collectionName, Map<String, String> optio
...
     } catch (SQLException e) {
-      LOGGER.error("Exception deleting table name: {}", collectionName);
+      LOGGER.error("Exception deleting table name: {}", tableIdentifier);
     }
     return false;
   }
 
   @Override
   public Collection getCollection(String collectionName) {
+    // FIXME - need to figure out listing schema tables before merging
```
kotharironak, line 145: Is it still pending? I see that we included the collections with schema as part of listCollections?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -98,16 +98,21 @@ public class PostgresCollection implements Collection {
...
   public PostgresCollection(final PostgresClient client, final String collectionName) {
+    this(client, PostgresTableIdentifier.parse(collectionName));
+  }
+
+  PostgresCollection(final PostgresClient client, final PostgresTableIdentifier tableIdentifier) {
     this.client = client;
-    this.collectionName = collectionName;
-    this.subDocUpdater = new PostgresSubDocumentUpdater(new PostgresQueryBuilder(collectionName));
-    this.queryExecutor = new PostgresQueryExecutor(collectionName);
+    this.tableIdentifier = tableIdentifier;
```
kotharironak, line 112: As `collectionName` is tied to the `Collection` object, if we could consider resolving collectionName here, we would only need to resolve it once, instead of resolving it every time it is accessed via the toString method of tableIdentifier. This is a very minor point, but I wanted to share it.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
aaron-steinfeld: > If the collectionName is 'x.y,' are we considering 'x' as the schema and 'y' as the table name? I see this in a PR comment too."

Yes

>Would there be a case of the collection name being 'x.y.z'? 

Could be. Mongo puts no restrictions. In this change this is now supported.

> In that case, the first part would be the schema, and 'y.z' would be the tableName. So, if we could support a dot as part of the table name, I think we should put them in the default schema public as we used to do before. This impl based logical grouping can lead to a higher number of random schemas and can caused backup/restore policy difficult in case of shared instance.

Initially went with this approach. The reason I didn't is that there are certain queries that do need to cross schemas. One of the bugs that motivated this change was the lack of postgres metrics, fixed in https://github.com/hypertrace/document-store/pull/181/commits/4dd0334965f700b1df6af4c6ef5eca67e8577a1e 

We could have introduced some postgres-specific way of specifying a schema separately from the collection I suppose, but I assumed our unit of backup/restore was the database rather than the schema - is that not the case?

> By the way, I had a thought of moving all collections under the 'hypertrace' schema, so it makes it easier for backup or any policies to apply if its used with shared instance with shared db. Well, for now, we are creating new db, so we can fully restore that db.

I guess you partially answer my last question here. What advantages would there be to using a schema over the current database approach? IMO, I like different services being able to maintain distinct schemas that are impl details.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
aaron-steinfeld: good catch! it's not, this is fixed now. will remove comment.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
aaron-steinfeld: Fair, but I'd prefer to keep the object form unless you object. Thinking - the overhead of toString is close to zero (basically an optional unwrap in the typical case or a string concat in the schema case), especially relative to the remote DB call that's part of the same sequence.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
kotharironak: > > If the collectionName is 'x.y,' are we considering 'x' as the schema and 'y' as the table name? I see this in a PR comment too."
> 
> Yes
> 
> > Would there be a case of the collection name being 'x.y.z'?
> 
> Could be. Mongo puts no restrictions. In this change this is now supported.
> 
> > In that case, the first part would be the schema, and 'y.z' would be the tableName. So, if we could support a dot as part of the table name, I think we should put them in the default schema public as we used to do before. This impl based logical grouping can lead to a higher number of random schemas and can caused backup/restore policy difficult in case of shared instance.
> 
> Initially went with this approach. The reason I ended up using the schema instead is that there are certain queries that do need to cross schemas. One of the bugs that motivated this change was the lack of postgres metrics, fixed in [4dd0334](https://github.com/hypertrace/document-store/commit/4dd0334965f700b1df6af4c6ef5eca67e8577a1e)
> 
> We could have introduced some postgres-specific way of specifying a schema separately from the collection I suppose, but I assumed our unit of backup/restore was the database rather than the schema - is that not the case?
> 
> > By the way, I had a thought of moving all collections under the 'hypertrace' schema, so it makes it easier for backup or any policies to apply if its used with shared instance with shared db. Well, for now, we are creating new db, so we can fully restore that db.
> 
> I guess you partially answer my last question here. What advantages would there be to using a schema over the current database approach? IMO, I like different services being able to maintain distinct schemas that are impl details.

Yes, currently, we are conducting maintenance at the database level. Agree that this approach allows us to logically group tables under a given database.

Since tables are created dynamically through the collection API, the corresponding schema is also created dynamically. Without a corresponding `deleteSchema API`, this could lead to dangling schemas and could introduce manual overhead.

As of now, we do not delete any collections except in test cases, so we can proceed. Later, we can extend the implementation of the `deleteCollection API` on the PostgreSQL side. If there are no tables left under the given non-public schema, we can then drop that schema as well.

kotharironak (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java
aaron-steinfeld: Makes sense. I considered the delete case too and skipped with similar thoughts.

PR #184, opened 2023-11-15
Introduce default keys for typesafe config extractor

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/model/config/TypesafeConfigDatastoreConfigExtractor.java
```diff
@@ -40,6 +53,20 @@ private TypesafeConfigDatastoreConfigExtractor(
     this.connectionPoolConfigBuilder = ConnectionPoolConfig.builder();
     this.datastoreConfigBuilder = DatastoreConfig.builder().type(type);
     this.endpointBuilder = Endpoint.builder();
+
+    final String dataStoreType = type.type();
+    this.hostKey(dataStoreType + "." + HOST_KEY)
```
aaron-steinfeld, line 58: If I'm not using one of these fields, will this cause an issue if they're missing? Basically want to make sure this is compatible if I'm not using auth, for example.

suresh-prakash: > will this cause an issue if they're missing?

Nope. If a config path is missing, the method invocation is a no-op.

aaron-steinfeld (approved):

skjindal93 (approved):

PR #186, opened 2023-11-30
make inclause behaviour consistent between mongo and postgres

SrikarMannepalli:


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -182,10 +186,28 @@ public static String parseNonCompositeFilterWithCasting(
       String fieldName, String columnName, String op, Object value, Builder paramsBuilder) {
     String parsedExpression =
         prepareCast(prepareFieldDataAccessorExpr(fieldName, columnName), value);
+    if (isInOp(op)) {
+      if (isFirstClassColumn(parsedExpression)) {
+        parsedExpression = "to_jsonb(" + parsedExpression + ")";
+      } else {
+        parsedExpression = parsedExpression.replace("->>", "->");
```
suresh-prakash, line 193: Wrapping with `PostgresFieldIdentifierExpressionVisitor` instead of `PostgresDataAccessorIdentifierExpressionVisitor` would directly give you in the format we need without explicit replace.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -182,10 +186,28 @@ public static String parseNonCompositeFilterWithCasting(
       String fieldName, String columnName, String op, Object value, Builder paramsBuilder) {
     String parsedExpression =
         prepareCast(prepareFieldDataAccessorExpr(fieldName, columnName), value);
+    if (isInOp(op)) {
+      if (isFirstClassColumn(parsedExpression)) {
+        parsedExpression = "to_jsonb(" + parsedExpression + ")";
```
suresh-prakash, line 191: Try to use a different RHS expression visitor that can use `to_jsonb(?)` or `?::jsonb` based on the type of the constant expression.
Ref.: `PostgresSubDocumentValueParser`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -352,6 +383,23 @@ public static String parseNonCompositeFilter(
...
+  private static StringBuilder prepareFilterStringForInOperator(
+      final String parsedExpression, final Iterable<Object> values, final Builder paramsBuilder) {
+    final StringBuilder filterStringBuilder = new StringBuilder();
+    filterStringBuilder.append("(");
+    final String collect =
+        StreamSupport.stream(values.spliterator(), false)
+            .map(
+                val -> {
+                  paramsBuilder.addObjectParam(val);
+                  return parsedExpression + " ?? ?";
```
suresh-prakash, line 395: I think, the parsedExpression will already contain `?`

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/PostgresQueryParserTest.java
```diff
@@ -158,16 +170,30 @@ void testParseNonCompositeFilterForJsonField() {
...
+      Filter filter = new Filter(Op.NOT_IN, "key1", List.of("abc"));
+      String query = PostgresQueryParser.parseFilter(filter, initParams());
+      Assertions.assertEquals("document->'key1' IS NULL OR NOT (document->'key1' ?? ?)", query);
     }
 
     {
       Filter filter = new Filter(Op.NOT_IN, "key1", List.of("abc", "xyz"));
       String query = PostgresQueryParser.parseFilter(filter, initParams());
-      Assertions.assertEquals("document->'key1' IS NULL OR document->>'key1' NOT IN (?, ?)", query);
+      Assertions.assertEquals(
```
suresh-prakash, line 194: With regards to document-store, it is better to start with a failing Integration test, followed by the fix and then the unit tests. Otherwise, the query may keep changing requiring modification of unit tests multiple times. 🙂

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -336,8 +365,10 @@ public static String parseNonCompositeFilter(
         throw new UnsupportedOperationException(UNSUPPORTED_QUERY_OPERATION);
     }
 
-    filterString.append(sqlOperator);
-    if (value != null) {
+    if (!isInOP) {
```
suresh-prakash, line 368: Can we check if we can simplify these conditions (or at least create different blocks)?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
SrikarMannepalli: Tried with integration tests, `?` seem to be there already.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
SrikarMannepalli: I am a bit confused as to whether we would need them as this would occur only in case of outer columns -> `id`, `created_at`, `last_updated_at`.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -352,6 +388,23 @@ public static String parseNonCompositeFilter(
...
+  private static StringBuilder prepareFilterStringForInOperator(
+      final String parsedExpression, final Iterable<Object> values, final Builder paramsBuilder) {
+    final StringBuilder filterStringBuilder = new StringBuilder();
+    filterStringBuilder.append("(");
+    final String collect =
+        StreamSupport.stream(values.spliterator(), false)
+            .map(
+                val -> {
+                  paramsBuilder.addObjectParam(val);
+                  return "(jsonb_build_array(" + parsedExpression + ") @> jsonb_build_array(?))";
```
suresh-prakash, line 400: Also, what happens in case when the LHS is a Function? It's fine if it's not supported. In case it is not supported, worth calling it out.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
SrikarMannepalli: Removed the old flow changes as per offline discussion.

SrikarMannepalli: Removed the old flow changes as per offline discussion.

suresh-prakash (commented):
Also, please add some integration tests.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -478,6 +505,10 @@ public static String prepareParsedNonCompositeFilter(
         throw new UnsupportedOperationException(UNSUPPORTED_QUERY_OPERATION);
     }
 
+    if (isInOP) {
+      return filterString.toString();
+    }
```
suresh-prakash, line 510: Could we simply return from the switch-case directly and avoid the boolean flag?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -434,13 +452,22 @@ public static String prepareParsedNonCompositeFilter(
         // NOTE: Pl. refer this in non-parsed expression for limitation of this filter
         sqlOperator = " NOT IN ";
         isMultiValued = true;
-        value = prepareParameterizedStringForList((Iterable<Object>) value, paramsBuilder);
+        isInOP = true;
+        filterString
+            .append(" IS NULL OR")
+            .append(" NOT ")
```
suresh-prakash, line 458: Does this `NOT` require wrapping around parentheses?
`NOT ...` or `NOT (...)`

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -352,6 +352,23 @@ public static String parseNonCompositeFilter(
...
+  private static StringBuilder prepareFilterStringForInOperator(
+      final String parsedExpression, final Iterable<Object> values, final Builder paramsBuilder) {
+    final StringBuilder filterStringBuilder = new StringBuilder();
+    filterStringBuilder.append("(");
+    final String collect =
+        StreamSupport.stream(values.spliterator(), false)
+            .map(
+                val -> {
+                  paramsBuilder.addObjectParam(val).addObjectParam(val);
+                  return "((jsonb_typeof(to_jsonb("+parsedExpression+")) = 'array' AND to_jsonb("+parsedExpression+") @> jsonb_build_array(?)) OR (jsonb_build_array(" + parsedExpression + ") @> jsonb_build_array(?)))";
```
suresh-prakash, line 364: Probably, spotless will fail on this line due to the missing space between the `+` sign.

document-store/src/integrationTest/resources/query/test_primitive_lhs_in_filter_aggr_response.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  {
+    "item":"Shampoo",
+    "quantity": 20
+  },
+  {
+    "item":"Shampoo",
+    "quantity": 10
+  }
+]
```
suresh-prakash, line 18: Nit: Empty line at the end of the file.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -361,7 +361,13 @@ private static StringBuilder prepareFilterStringForInOperator(
             .map(
                 val -> {
                   paramsBuilder.addObjectParam(val).addObjectParam(val);
-                  return "((jsonb_typeof(to_jsonb("+parsedExpression+")) = 'array' AND to_jsonb("+parsedExpression+") @> jsonb_build_array(?)) OR (jsonb_build_array(" + parsedExpression + ") @> jsonb_build_array(?)))";
+                  return "((jsonb_typeof(to_jsonb("
```
suresh-prakash, line 364: Apologies for not catching it before. Can use `String.format()` and substitute placeholders of `%s` to improve readability.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -450,25 +455,31 @@ public static String prepareParsedNonCompositeFilter(
       case "NOT_IN":
       case "NOT IN":
         // NOTE: Pl. refer this in non-parsed expression for limitation of this filter
+        // In order to make the behaviour same as for mongo, the not in operator would match only if
+        // the lhs
+        // and rhs have no intersection at all
+        // NOTE: This doesn't work in case the lhs is a function
```
suresh-prakash, line 461: Nit:
```suggestion
        // In order to make the behaviour same as for Mongo, the "NOT_IN" operator would match only if
        // the LHS and RHS have no intersection at all
        // NOTE: This doesn't work in case the LHS is a function
```

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/utils/PostgresUtils.java
```diff
@@ -450,25 +455,31 @@ public static String prepareParsedNonCompositeFilter(
...
-                    preparedExpression, (Iterable<Object>) value, paramsBuilder));
-        break;
+                    preparedExpression, (Iterable<Object>) value, paramsBuilder))
+            .append(")");
+        return filterString.toString();
       case "IN":
+        // In order to make the behaviour same as for mongo, the in operator would match if the lhs
+        // and rhs have any intersection at all
         // NOTE: Pl. refer this in non-parsed expression for limitation of this filter
+        // NOTE: This doesn't work in case the lhs is a function
```
suresh-prakash, line 476:
```suggestion
        // In order to make the behaviour same as for Mongo, the "INI" operator would match if the 
        // LHS and RHS have any intersection (i.e. non-empty intersection)
        // NOTE: Pl. refer this in non-parsed expression for limitation of this filter
        // NOTE: This doesn't work in case the LHS is a function
```

suresh-prakash:
Only nits. Overall, looks good.

suresh-prakash (approved):

PR #187, opened 2023-12-12
Handle the zero limit case explicitly for MongoDB aggregation

suresh-prakash:
MongoDB `find()` implementation handles the 0-limit case and returns no documents.
Whereas the deprecation alternative method `aggregate()` implementation throws an exception on a 0-limit for MongoDB, while it works as expected for Postgres.

This PR aims at fixing that bug by manually handling the 0-limit case (as there is no way to handle this in MongoDB aggregation pipeline) and adds an integration test to cover that scenario.


kotharironak (approved):

sarthak77:
>MongoDB find() implementation handles the 0-limit case and returns no documents.

Does it return no documents or all 🤔

PR #192, opened 2024-01-07
Introduce NOT logical operator and add tests

suresh-prakash:
Introduce the 'NOT' logical operator for both MongoDB and Postgres.

Note: 
When combined with array operator (`ANY`), this can help achieve the `NONE` array operator because 
```
NONE = NOT ANY
```

Also, with double negation, we can achieve the `ALL` array operator.
```
ALL meet a condition = NOT ANY that does NOT meet the condition
```

Integration tests are added to capture these scenarios.


document-store/src/main/java/org/hypertrace/core/documentstore/model/config/mongo/MongoConnectionConfig.java
```diff
@@ -140,7 +140,7 @@ private void applyClusterSettings(final Builder settingsBuilder) {
   private void applyConnectionPoolSettings(final Builder settingsBuilder) {
     final ConnectionPoolSettings connectionPoolSettings =
         ConnectionPoolSettings.builder()
-            .maxConnecting(connectionPoolConfig.maxConnections())
+            .maxSize(connectionPoolConfig.maxConnections())
```
kotharironak, line 143: Would impact any existing client?

suresh-prakash: This was a bug to be fixed.

kotharironak (approved):

PR #197, opened 2024-05-20
Enable disk usage for aggregations on MongoDB

suresh-prakash:
Some of the aggregation queries could fail because of memory limits per aggregation pipeline on MongoDB. To avoid such issue, this PR allows disk usage so that the queries that couldn't fit in the memory limits can leverage the disk usage.

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/MongoQueryExecutor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/MongoQueryExecutor.java
@@ -130,7 +130,8 @@ public class MongoQueryExecutor {
     logPipeline(pipeline);
 
     try {
-      final AggregateIterable<BasicDBObject> iterable = collection.aggregate(pipeline);
+      final AggregateIterable<BasicDBObject> iterable =
+          collection.aggregate(pipeline).allowDiskUse(true);
       return iterable.cursor();
     } catch (final MongoCommandException e) {
       log.error("Execution failed for query: {}. Aggregation Pipeline: {}", query, pipeline);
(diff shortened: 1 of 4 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/MongoQueryExecutor.java
```diff
@@ -130,7 +130,8 @@ public MongoCursor<BasicDBObject> aggregate(final Query originalQuery) {
     logPipeline(pipeline);
 
     try {
-      final AggregateIterable<BasicDBObject> iterable = collection.aggregate(pipeline);
+      final AggregateIterable<BasicDBObject> iterable =
+          collection.aggregate(pipeline).allowDiskUse(true);
```
pavan-traceable, line 134: In what scenarios it will use disk ? Depends on size of data or by default it uses disk  ?

suresh-prakash: > Use [allowDiskUse()](https://www.mongodb.com/docs/manual/reference/method/cursor.allowDiskUse/?_ga=2.154934904.1773902492.1716202636-852641550.1716202636#mongodb-method-cursor.allowDiskUse) to either allow or prohibit writing temporary files on disk when a pipeline stage exceeds the 100 megabyte limit. Starting in MongoDB 6.0, operations that require greater than 100 megabytes of memory automatically write data to temporary files by default.

Ref.: https://www.mongodb.com/docs/manual/reference/method/cursor.allowDiskUse/?_ga=2.154934904.1773902492.1716202636-852641550.1716202636#mongodb-method-cursor.allowDiskUse

aman-bansal (approved):

aman-bansal (approved):

skjindal93 (approved):

PR #207, opened 2024-08-29
Handled dollar prefixed values for "createOrReplace" Mongo implementation

suresh-prakash:


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoUtils.java
```diff
@@ -113,6 +124,17 @@ public static JsonNode recursiveClone(
       if (value.isObject()) {
         newValue = recursiveClone(value, function, emptyObjectConverter);
       }
+      if (value.isArray()) {
```
aaron-steinfeld, line 127: Typically in this kind of function you'd have object/array/primitives as sibling, with recursion inside object and array. Here it looks like we handle primitives, else assume it's an object and only treat array differently inside the object fields? Am I reading that right and if so, is there a reason to do it that way?

suresh-prakash: We are treating "primitive values" differently. 
Also, we are treating "object keys" differently.
Both needs encoding so that a `$` prefix is (treated literally and) not treated as the special field lookup operator.

We are still doing similar thing. Nested inside objects and arrays and handling primitives differently. Because, we also want object keys to be encoded, this method turned out this way. 

In this PR, I'm just handling the missing case of "primitive values" (especially when they are strings). So, just piggy bagging onto the existing method.

aaron-steinfeld (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoUtils.java
```diff
@@ -113,6 +124,17 @@ public static JsonNode recursiveClone(
...
+        newValue = new ArrayNode(JsonNodeFactory.instance);
+        for (int i = 0; i < value.size(); i++) {
+          final JsonNode elementValue =
+              recursiveClone(value.get(i), function, emptyObjectConverter);
+          ((ArrayNode) newValue).add(elementValue);
+        }
+      }
+      if (newValue.isTextual()) {
+        newValue = new TextNode(function.apply(newValue.asText()));
```
kotharironak, line 136: Why do we need this, we have already applied transformation to each value?

kotharironak: Sorry my mistake. I read it as array node.

kotharironak (approved):

PR #209, opened 2024-09-13
Introduced data freshness with the MongoDB implementation at connection-level and at query-level

suresh-prakash:
This PRs gives the flexibility to redirect reads optionally to secondary MongoDB nodes. 
This configuration can be tweaked at the connection-level while instantiating a connection-config object. It can also be configured at a query level. By default, all the reads will go to the primary node for backward compatibility.

Preference order:
1. Query level configuration
2. Connection level configuration if query level configuration is default (unspecified)
3. Document store defaults to reads from primary (if both query-level and connection-level read preferences are unspecified/default).

In this PR, the Postgres implementation simply ignores the configuration.


build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
   id("org.hypertrace.ci-utils-plugin") version "0.3.0"
   id("org.hypertrace.publish-plugin") version "1.0.2" apply false
   id("org.hypertrace.code-style-plugin") version "1.1.0" apply false
-  id("org.owasp.dependencycheck") version "8.2.1"
+  id("org.owasp.dependencycheck") version "8.4.3"
```
kotharironak, line 9: Any reason for not moving to latest version - `10.0.4`?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreTest.java
```diff
@@ -71,7 +71,7 @@ public class DocStoreTest {
   public static void init() {
     datastoreMap = Maps.newHashMap();
     mongo =
-        new GenericContainer<>(DockerImageName.parse("mongo:4.4.0"))
+        new GenericContainer<>(DockerImageName.parse("mongo:7.0.14"))
```
kotharironak, line 74: Is the MongoDB version 7.0.14? Is this the same version that has been deployed?

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -127,6 +128,16 @@ public interface Collection {
...
   CloseableIterator<Document> aggregate(final org.hypertrace.core.documentstore.query.Query query);
 
+  /**
+   * Query the documents conforming to the query specification.
+   *
+   * @param query The query specification
+   * @param queryOptions The query options
+   * @return {@link CloseableIterator} of matching documents
+   */
+  CloseableIterator<Document> query(
```
kotharironak, line 138: Q: the existing query api deprecated?

document-store/src/main/java/org/hypertrace/core/documentstore/model/options/DataFreshness.java
```diff
@@ -0,0 +1,9 @@
+package org.hypertrace.core.documentstore.model.options;
+
+@SuppressWarnings("UnnecessarySemicolon")
+public enum DataFreshness {
+  SYSTEM_DEFAULT,
+  REALTIME_FRESHNESS,
```
kotharironak, line 6: Q: What does the freshness mean?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/collection/MongoCollectionOptionsApplier.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public MongoCollection<BasicDBObject> applyOptions(
+      final ConnectionConfig connectionConfig,
+      final QueryOptions queryOptions,
+      final MongoCollection<BasicDBObject> collection) {
+    final CacheKey cacheKey =
+        CacheKey.builder().readPreference(readPreference(connectionConfig, queryOptions)).build();
+    return collectionCache.computeIfAbsent(
+        cacheKey, key -> collection.withReadPreference(key.readPreference()));
```
kotharironak, line 31: How is the CacheKey associated with the collection? Since ReadPreference can be either primary or secondary, does this mean that `this.collectionCache` contains a maximum of 2 values?

build.gradle.kts
suresh-prakash: I tried that version. It's failing (mostly due to Java version incompatibility).

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
suresh-prakash: Don't think that's required. If the clients want to query with the default configurations, the older method can be used.

document-store/src/main/java/org/hypertrace/core/documentstore/model/options/DataFreshness.java
suresh-prakash: Freshness means the liveliness of the data.
REALTIME_FRESHNESS means we want the most up-to-date data.
NEAR_REALTIME_FRESHNESS means a little bit of stale data is acceptable.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/collection/MongoCollectionOptionsApplier.java
suresh-prakash: > Since ReadPreference can be either primary or secondary, does this mean that this.collectionCache contains a maximum of 2 values?

Yes.

document-store/src/main/java/org/hypertrace/core/documentstore/model/options/DataFreshness.java
kotharironak: Here `NEAR_REALTIME_FRESHNESS` means fetching data from secondary, right?

kotharironak (approved):

document-store/src/main/java/org/hypertrace/core/documentstore/model/options/DataFreshness.java
suresh-prakash: Yes. Basically "SECONDARY" is preferred. In case "SECONDARY" is not available, it'll fetch from "PRIMARY".

PR #211, opened 2024-09-17
Introduced query timeout parameter

suresh-prakash:
Introduced query timeout parameter

* If unspecified, a cluster-level default value of 20 minutes will be used as the timeout
* If unspecified, a query-level default value of 1 minute will be used as the timeout
* The Postgres implementation simply continues to ignore the query options
* Minimum of cluster-level value and the query-level value will be used
* Redirect the custom metric reporting to secondary nodes always
* Use a query-level timeout override of 20 minutes for reporting custom metrics


document-store/src/main/java/org/hypertrace/core/documentstore/TypesafeDatastoreConfigAdapter.java
```diff
@@ -32,7 +33,8 @@ public DatastoreConfig convert(final Config config) {
               null,
               null,
               AggregatePipelineMode.DEFAULT_ALWAYS,
-              DataFreshness.SYSTEM_DEFAULT) {
+              DataFreshness.SYSTEM_DEFAULT,
+              Duration.ofMinutes(20)) {
```
kotharironak, line 37: What does 20 mins mean here?

document-store/src/main/java/org/hypertrace/core/documentstore/metric/BaseDocStoreMetricProviderImpl.java
```diff
@@ -55,63 +58,70 @@ public List<DocStoreMetric> getCustomMetrics(final CustomMetricConfig customMetr
...
 
-      while (iterator.hasNext()) {
-        final Document document = iterator.next();
-        final JsonNode node;
+      try (final CloseableIterator<Document> iterator =
+          collection.query(
+              query,
+              QueryOptions.builder()
+                  .dataFreshness(NEAR_REALTIME_FRESHNESS)
+                  .queryTimeout(Duration.ofMinutes(20))
```
kotharironak, line 68: Is there a way to expose queryTime out as config?

kotharironak (approved):
LGTM

document-store/src/main/java/org/hypertrace/core/documentstore/TypesafeDatastoreConfigAdapter.java
suresh-prakash: Default query timeout in case no timeout is specified by the clients.

document-store/src/main/java/org/hypertrace/core/documentstore/metric/BaseDocStoreMetricProviderImpl.java
suresh-prakash: Yes. It's also available at the "ConnectionConfig" level, which can be overridden here if required.

PR #220, opened 2025-04-01
feat: add support for self-joins with sub-query

GurtejSohi:
## Description
This PR has changes to add support for self-joins with sub-query (represented by `SubQueryJoinExpression`) in MongoDB. 
Note that postgres impl of `FromTypeExpressionVisitor` for visiting `SubQueryJoinExpression` throws an `UnsupportedOperationException`. Currently, only the mongo implementation supports this operation.

### Testing
Added an integration test for testing the mongo impl of self-join with sub-query. It also demonstrates its sample usage


document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/JoinExpression.java
```diff
@@ -0,0 +1,38 @@
...
+import org.hypertrace.core.documentstore.parser.SelectTypeExpressionVisitor;
+import org.hypertrace.core.documentstore.query.Query;
+
+/**
+ * Expression representing a condition for filtering
+ *
+ * <p>Example: <code>
+ * company IN ('Traceable', 'Harness')
+ * </code> can be constructed as <code>
+ * RelationalExpression.of( IdentifierExpression.of("company"), RelationalOperator.IN,
```
suresh-prakash, line 20: Seems like copy-paste error.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/JoinExpression.java
```diff
@@ -0,0 +1,38 @@
...
+ * ConstantExpression.ofStrings("Traceable", "Harness"))));
+ * </code>
+ */
+@Value
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class JoinExpression implements FromTypeExpression {
+  FromTypeExpression left;
+  FromTypeExpression right;
+  JoinType joinType;                    // INNER, LEFT, RIGHT, etc.
```
suresh-prakash, line 30: We may not need the join type unless it is required for the current use case.

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/JoinExpression.java
```diff
@@ -0,0 +1,38 @@
...
+ * </code> can be constructed as <code>
+ * RelationalExpression.of( IdentifierExpression.of("company"), RelationalOperator.IN,
+ * ConstantExpression.ofStrings("Traceable", "Harness"))));
+ * </code>
+ */
+@Value
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class JoinExpression implements FromTypeExpression {
+  FromTypeExpression left;
```
suresh-prakash, line 28: Wouldn't this be implicit?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/SubQueryFromExpression.java
```diff
@@ -0,0 +1,26 @@
...
+import org.hypertrace.core.documentstore.parser.FromTypeExpressionVisitor;
+import org.hypertrace.core.documentstore.query.Query;
+
+/**
+ * Allows embedding an entire Query in the FROM clause.
+ */
+@Value
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class SubQueryFromExpression implements FromTypeExpression {
```
suresh-prakash, line 17: We would support either this or `JoinExpression`, right?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/TableFromExpression.java
```diff
@@ -0,0 +1,24 @@
...
+import org.hypertrace.core.documentstore.expression.type.FromTypeExpression;
+import org.hypertrace.core.documentstore.parser.FromTypeExpressionVisitor;
+
+/**
+ * A simple wrapper for referencing a physical collection/table in the FROM clause.
+ */
+@Value
+@Builder(toBuilder = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class TableFromExpression implements FromTypeExpression {
```
suresh-prakash, line 16: Do we also need this?

document-store/src/test/java/org/hypertrace/core/documentstore/expression/impl/JoinExpressionTest.java
```diff
@@ -0,0 +1,107 @@
...
+        .addSelection(
+            SelectionSpec.of(IdentifierExpression.of("suite_id"), "suite_id")
+        )
+        .addSelection(
+            SelectionSpec.of(
+                AggregateExpression.of(AggregationOperator.MAX, IdentifierExpression.of("scan_run_number")),
+                "latest_scan_run_number"
+        ))
+        // FROM
+        .addFromClause(
```
suresh-prakash, line 44: We've always had this implicit. Is it required now?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/JoinExpression.java
GurtejSohi: yes, fixed it

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/SubQueryFromExpression.java
GurtejSohi: Typically, this will be the `right` field in `JoinExpression`, as shown in the example in `JoinExpressionTest`

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/TableFromExpression.java
GurtejSohi: updated the API, so this is outdated

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/AliasedIdentifierExpression.java
```diff
@@ -0,0 +1,30 @@
...
+@Value
+public class AliasedIdentifierExpression extends IdentifierExpression {
+  String alias;
+
+  private AliasedIdentifierExpression(final String name, final String alias) {
+    super(name);
+    this.alias = alias;
+  }
+
+  public static AliasedIdentifierExpression of(final String name, final String alias) {
```
suresh-prakash, line 20: Direct/static constructors like this with multiple arguments of the same datatype are error-prone since the callers can accidentally pass them in incorrect order.

Please prefer a builder instead.

document-store/src/test/java/org/hypertrace/core/documentstore/expression/impl/SubQueryJoinExpressionTest.java
```diff
@@ -0,0 +1,82 @@
...
+    SubQueryJoinExpression subQueryJoinExpression =
+        SubQueryJoinExpression.builder()
+            .subQuery(subQuery)
+            .subQueryAlias("latest")
+            .joinCondition(
+                LogicalExpression.and(
+                    RelationalExpression.of(
+                        IdentifierExpression.of("suite_id"),
+                        RelationalOperator.EQ,
+                        IdentifierExpression.of("latest.suite_id")),
```
suresh-prakash, line 61: Shouldn't this be AliasedIdentifierExpression?

document-store/src/test/java/org/hypertrace/core/documentstore/expression/impl/SubQueryJoinExpressionTest.java
```diff
@@ -0,0 +1,82 @@
...
+                    RelationalExpression.of(
+                        IdentifierExpression.of("scan_run_number"),
+                        RelationalOperator.EQ,
+                        IdentifierExpression.of("latest.latest_scan_run_number"))))
```
suresh-prakash, line 65: Shouldn't this be AliasedIdentifierExpression?

document-store/src/test/java/org/hypertrace/core/documentstore/expression/impl/SubQueryJoinExpressionTest.java
GurtejSohi: yes, will update it

document-store/src/test/java/org/hypertrace/core/documentstore/expression/impl/SubQueryJoinExpressionTest.java
GurtejSohi: yes

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+                .rhsParser(
+                    new MongoDollarPrefixingIdempotentParser(
+                        new MongoAliasedIdentifierExpressionParser()))
+                .build());
+    final Map<String, Object> filter = joinCondition.accept(parser);
+    return new BasicDBObject(filter);
+  }
+
+  private String getJoinedResultFieldName(String subQueryAlias) {
+    return "__joined_result_with_" + subQueryAlias;
```
suresh-prakash, line 134: Could you please explain what this is for?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+
+    unwindStage.put(UNWIND_OPERATOR, unwindSpec);
+    return unwindStage;
+  }
+
+  private BasicDBObject createReplaceRootStage(String joinedResultFieldName) {
+    BasicDBObject replaceRootStage = new BasicDBObject();
+    BasicDBObject newRoot = new BasicDBObject();
+
+    newRoot.put("newRoot", MongoUtils.PREFIX + joinedResultFieldName);
```
suresh-prakash, line 152: Nit: Prefer static constants.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+
+  MongoLetClauseBuilder(String subQueryAlias) {
+    this.subQueryAlias = subQueryAlias;
+  }
+
+  @Override
+  public BasicDBObject visit(LogicalExpression expression) {
+    BasicDBObject letClause = new BasicDBObject();
+    for (FilterTypeExpression operand : expression.getOperands()) {
+      letClause.putAll((Map<String, Object>) operand.accept(this));
```
suresh-prakash, line 32: We seem to be casting `BasicDBObject` into a `Map`. This could be confusing (probably compiler might be generating a warning).

Can we either convert everything to `BasicDBObject` or everything to `Map`?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private class MongoLetClauseSelectTypeExpressionVisitor implements SelectTypeExpressionVisitor {
+    @Override
+    public BasicDBObject visit(AggregateExpression expression) {
+      return new BasicDBObject();
+    }
+
+    @Override
+    public BasicDBObject visit(ConstantExpression expression) {
+      return new BasicDBObject();
```
suresh-prakash, line 57: Is this deliberate? 
If this is not expected here, why not extend the `MongoSelectTypeExpressionParser`?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public BasicDBObject visit(DocumentArrayFilterExpression expression) {
+    return new BasicDBObject();
+  }
+
+  private BasicDBObject createLetClause(
+      AliasedIdentifierExpression aliasedExpression, String subQueryAlias) {
+    BasicDBObject letClause = new BasicDBObject();
+    if (aliasedExpression.getContextAlias().equals(subQueryAlias)) {
+      letClause.put(aliasedExpression.getName(), "$" + aliasedExpression.getName());
```
suresh-prakash, line 100: What happens if the `AliasedExpression` contains a nested field `like "a.b"`. 
Does it need escaping/encoding? 

It would be beneficial to add an integration for the same.

document-store/src/main/java/org/hypertrace/core/documentstore/parser/FromTypeExpressionVisitor.java
```diff
@@ -1,7 +1,15 @@
...
 import org.hypertrace.core.documentstore.expression.impl.UnnestExpression;
 
 public interface FromTypeExpressionVisitor {
   <T> T visit(UnnestExpression unnestExpression);
+
+  /*
+   * Subquery join expression is not supported by default. Override this method to support it.
+   */
+  default <T> T visit(SubQueryJoinExpression subQueryJoinExpression) {
+    throw new UnsupportedOperationException("This operation is not supported");
```
suresh-prakash, line 13: Typically, we should force all the implementors of visitor interface to implement all the methods.

Whether to throw exception or not should be the implementor's responsibility. 

We probably created the base class 
 `MongoSelectTypeExpressionParser` for a similar purpose (without having a default implementation in the interface).

The reason being, it may make sense to implement this in other implementations (other than MongoDB).

document-store/src/main/java/org/hypertrace/core/documentstore/parser/SelectTypeExpressionVisitor.java
```diff
@@ -16,4 +17,8 @@ public interface SelectTypeExpressionVisitor {
   <T> T visit(final FunctionExpression expression);
 
   <T> T visit(final IdentifierExpression expression);
+
+  default <T> T visit(final AliasedIdentifierExpression expression) {
+    throw new UnsupportedOperationException();
```
suresh-prakash, line 22: Same here. Not a typical visitor flow.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: yes, this is deliberate as this visitor is expected to go through SelectTypeExpressions and only return the fields that have to be declared in the let clause. So any other field other than AliasedIdentifierExpression has to be ignored but it should not throw an error in that case and for the case of AliasedIdentifierExpression it should return the field name that we need to add to the let clause.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: The field name of the aliasedExpression is given by the user and if we can make an assumption that the user won't give a dot in the field name then we should be fine as we are only using the field name here and not the context alias.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
suresh-prakash: Got it. This may be a common use-case. So, can we create a common base class that returns default (empty map/BasicDBObject) for every implementation and then override it? 

That way
1. The default returning shared base-class can be re-used
2. Implementation here would look concise

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
suresh-prakash: I'm a bit confused. Does it allow me to pick the value of "a.b" and alias it as "c"? 

In other words, are you saying that the alias can't contain a dot, but the identifier name can?

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/AliasedIdentifierExpression.java
```diff
@@ -0,0 +1,69 @@
...
+ *         GROUP BY item
+ *     ) AS latest
+ *     ON item = latest.item
+ *     ORDER BY `item` ASC;
+ * </code> the rhs of the join condition "latest.item" can be expressed as: <code>
+ *  AliasedIdentifierExpression.builder().name("item").alias("alias1").build() </code>
+ */
+@Value
+public class AliasedIdentifierExpression extends IdentifierExpression {
+  String contextAlias;
```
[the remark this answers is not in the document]
GurtejSohi: Alias sounds like we are giving the field name as alias (like in [SelectionSpec](https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/query/SelectionSpec.java#L20). If you look at the example above, this is not the alias for the field, but the alias for the context in which we are referring the field.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
GurtejSohi: This is the field name in which the joined results will be put after the lookup stage. This field name can be used by subsequent stages in the aggregate pipeline to do operations on the joined result.
^Also added it as a comment

document-store/src/integrationTest/resources/query/self_join_with_sub_query_response.json
```diff
@@ -0,0 +1,6 @@
+[
+  {"date":"2015-09-10T08:43:00Z","item":"Comb","quantity":10},
+  {"date":"2014-03-01T09:00:00Z","item":"Mirror","quantity":1},
+  {"date":"2014-04-04T11:21:39.736Z","item":"Shampoo","quantity":20},
+  {"date":"2016-02-06T20:20:13Z","item":"Soap","quantity":5}
+]
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/integrationTest/resources/query/self_join_with_sub_query_response.json
```diff
@@ -0,0 +1,6 @@
+[
+  {"date":"2015-09-10T08:43:00Z","item":"Comb","quantity":10},
+  {"date":"2014-03-01T09:00:00Z","item":"Mirror","quantity":1},
+  {"date":"2014-04-04T11:21:39.736Z","item":"Shampoo","quantity":20},
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/expression/impl/SubQueryJoinExpression.java
```diff
@@ -0,0 +1,36 @@
...
+  FilterTypeExpression joinCondition;
+
+  @Override
+  public <T> T accept(FromTypeExpressionVisitor visitor) {
+    return visitor.visit(this);
+  }
+
+  @Override
+  public String toString() {
+    return String.format("JOIN (%s) AS %s ON %s", subQuery, subQueryAlias, joinCondition);
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+  @SuppressWarnings("unchecked")
+  @Override
+  public List<BasicDBObject> visit(SubQueryJoinExpression subQueryJoinExpression) {
+    this.mongoLetClauseBuilder =
+        new MongoLetClauseBuilder(subQueryJoinExpression.getSubQueryAlias());
+
+    Query transformedSubQuery =
+        MongoQueryTransformer.transform(subQueryJoinExpression.getSubQuery());
+    List<BasicDBObject> aggregatePipeline =
+        new ArrayList<>(mongoQueryExecutor.convertToAggregatePipeline(transformedSubQuery));
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+    aggregatePipeline.add(createLookupStage(subQueryJoinExpression, joinedResultFieldName));
+
+    // Lookup Stage puts the joined results into an array field. We need to unwind that array field
+    // to get the joined results as separate documents.
+    aggregatePipeline.add(createUnwindStage(joinedResultFieldName));
+
+    // Replace root with the joined document
+    aggregatePipeline.add(createReplaceRootStage(joinedResultFieldName));
+
+    return aggregatePipeline;
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+  }
+
+  private BasicDBObject createLookupStage(
+      SubQueryJoinExpression subQueryJoinExpression, String joinedResultFieldName) {
+    BasicDBObject lookupStage = new BasicDBObject();
+    BasicDBObject lookupSpec = new BasicDBObject();
+
+    lookupSpec.put("from", mongoQueryExecutor.getCollectionName());
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+    lookupSpec.put("as", joinedResultFieldName);
+
+    lookupStage.put(LOOKUP_OPERATOR, lookupSpec);
+    return lookupStage;
+  }
+
+  private List<BasicDBObject> createLookupPipeline(SubQueryJoinExpression subQueryJoinExpression) {
+    List<BasicDBObject> pipeline = new ArrayList<>();
+    pipeline.add(createMatchStage(subQueryJoinExpression));
+    return pipeline;
```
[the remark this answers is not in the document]
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
```diff
@@ -42,9 +61,103 @@ public List<BasicDBObject> visit(UnnestExpression unnestExpression) {
...
+  private BasicDBObject getFilterClause(FilterTypeExpression joinCondition) {
+    final FilterTypeExpressionVisitor parser =
+        new MongoFilterTypeExpressionParser(
+            MongoRelationalFilterContext.builder()
+                .location(FilterLocation.INSIDE_EXPR)
+                .lhsParser(
+                    new MongoDollarPrefixingIdempotentParser(new MongoIdentifierExpressionParser()))
+                .rhsParser(
+                    new MongoDollarPrefixingIdempotentParser(
+                        new MongoAliasedIdentifierExpressionParser()))
```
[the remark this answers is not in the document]
GurtejSohi: yes, earlier we were enforcing the order but we can avoid it. Updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoFromTypeExpressionParser.java
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: changed the return type to map

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: Sorry, my bad, I misunderstood your previous comment. I went through it again and it made sense this time. I've updated the implementation to encode the variable name so that it can handle nested fields also. I've also added an integration test for it.

document-store/src/main/java/org/hypertrace/core/documentstore/parser/FromTypeExpressionVisitor.java
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/parser/SelectTypeExpressionVisitor.java
GurtejSohi: updated it

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+    return Collections.emptyMap();
+  }
+
+  @Override
+  public Map<String, Object> visit(ArrayRelationalFilterExpression expression) {
+    return Collections.emptyMap();
+  }
+
+  @Override
+  public Map<String, Object> visit(DocumentArrayFilterExpression expression) {
```
suresh-prakash, line 67: We don't need to override these, right?

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoUtils.java
```diff
@@ -65,6 +65,13 @@ public static String decodeKey(final String key) {
     return key.replace("\\u002e", FIELD_SEPARATOR).replace("\\u0024", PREFIX).replace("\\\\", "\\");
   }
 
+  public static String encodeVariableName(final String variableName) {
```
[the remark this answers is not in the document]
GurtejSohi: Encode method in MongoUtils also adds backslashes. Backslashes are allowed in keys but not in variable names, as mentioned here - https://www.mongodb.com/docs/manual/reference/aggregation-variables/

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/MongoLetClauseBuilder.java
GurtejSohi: I just realized that it might be confusing because I put the inner class in between the outer class methods. 
Did you think that these are methods of the inner class? If so, please note that these are outer class methods and since it implements filter type expression it has to implement these methods as well.

To avoid confusion, I've moved the inner class below the outer class methods

suresh-prakash: Oh, yeah. That was the confusion. Makes sense to keep the inner classes towards the end.

suresh-prakash (approved):

PR #226, opened 2025-07-15
fix: update java compatibility

aaron-steinfeld:


suresh-prakash (approved):

build.gradle.kts
```diff
@@ -6,6 +6,7 @@ plugins {
   id("org.hypertrace.ci-utils-plugin") version "0.4.0"
   id("org.hypertrace.publish-plugin") version "1.1.1" apply false
   id("org.hypertrace.code-style-plugin") version "2.1.2" apply false
+  id("org.hypertrace.java-convention") version "0.4.0"
```
aaron-steinfeld, line 9: Our java convention plugin decouples the build from the current java version. By default, it uses java 21 to compile but targets compatibility with 11. The current issue is that it's using whatever version was used to run gradle, and CI's default version was updated to 17 by GH.

document-store/build.gradle.kts
```diff
@@ -22,7 +22,6 @@ dependencies {
 
   testImplementation(libs.org.junit.jupiter.junit.jupiter)
   testImplementation(libs.org.mockito.mockito.core)
-  testImplementation(libs.org.mockito.mockito.inline)
```
aaron-steinfeld, line 25: Newer versions of mockito have this built in. The old version was not compatible with newer java.

PR #232, opened 2025-09-09
Adding  logic and fixing order

Sukhamjot-Singh:
## Description
There a was a bug when using NOT_IN in the array pipeline.

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/filter/MongoStandardExprNotInRelationalFilterParser.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/filter/MongoStandardExprNotInRelationalFilterParser.java
@@ -1,0 +1,26 @@
+package org.hypertrace.core.documentstore.mongo.query.parser.filter;
+
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import org.hypertrace.core.documentstore.expression.impl.RelationalExpression;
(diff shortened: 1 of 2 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/mongo/query/parser/filter/MongoStandardExprNotInRelationalFilterParser.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.AllArgsConstructor;
+import org.hypertrace.core.documentstore.expression.impl.RelationalExpression;
+import org.hypertrace.core.documentstore.mongo.query.parser.filter.MongoRelationalFilterParserFactory.MongoRelationalFilterContext;
+
+/**
+ * Note: MongoDB does not have a native "NOT IN" aggregation operator. This parser simulates "NOT
+ * IN" functionality by combining the $not and $in operators.
+ */
+@AllArgsConstructor
+public class MongoStandardExprNotInRelationalFilterParser implements MongoRelationalFilterParser {
```
suresh-prakash, line 13: My bad on the naming suggestion earlier. Since we are special casing "NOT_IN", it is no longer a standard one.

```suggestion
public class MongoNotInExprRelationalFilterParser implements MongoRelationalFilterParser {
```

Sukhamjot-Singh:
> You can check-in your integration tests (in a generic way without specifying any proprietary information) so that any changes to this flow would be caught early on.

I have added test for Mongo provider
I couldn't debug postgres issue

PR #288, opened 2026-04-12
Add Collection#bulkUpdate(Map<Key, Collection<SubDocumentUpdate>> updates, ..)

suddendust:
## Description
Since the older API `Collection#bulkUpdateSubDocs` is deprecated, we did not have any new method to update multiple records/docs at with its own set of updates. This change introduces this API. Also, unlike the older, deprecated API that supported only `SET`, this new API supports all 5 sub-doc operators.

### Testing
Added integration tests

### Checklist:

### Documentation
This method does not guarantee any atomicity. It means partial updates are allowed. `FlatPostgresCollection#bulkUpdateSubDocs` is not atomic as well, as per the contract.


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
```diff
@@ -896,6 +907,104 @@ public CloseableIterator<Document> bulkUpdate(
...
+    for (Map.Entry<Key, Collection<SubDocumentUpdate>> entry : updates.entrySet()) {
+      Key key = entry.getKey();
+      Collection<SubDocumentUpdate> keyUpdates = entry.getValue();
+
+      if (keyUpdates == null || keyUpdates.isEmpty()) {
+        continue;
+      }
+
+      try {
+        boolean updated = updateSingleKey(key, keyUpdates, tableName, quotedPkColumn);
```
puneet-traceable: I see that this method tries getting a connection from pool per query. would it make sense to fire this on the same connection?

suddendust: Oh, you mean use the same connection for the entire batch? Yes, that's a good optimisation. Will change this.

suddendust: This is done

document-store/src/main/java/org/hypertrace/core/documentstore/Collection.java
```diff
@@ -398,5 +398,53 @@ CloseableIterator<Document> bulkUpdate(
...
       throws IOException;
 
+  /**
+   * Bulk update sub-documents with key-specific updates. Each key can have its own set of
+   * SubDocumentUpdate operations, allowing different updates per document.
+   *
+   * <p>This method supports all update operators (SET, UNSET, ADD, APPEND_TO_LIST,
+   * ADD_TO_LIST_IF_ABSENT, REMOVE_ALL_FROM_LIST). Updates for each individual key are applied
+   * atomically, but there is no atomicity guarantee across different keys - some keys may be
+   * updated while others fail. Any atomicity guarantees are implementation-specific.
```
suresh-prakash: > Any atomicity guarantees are implementation-specific.

This violates the Liskov's substitution principle ('L' in SOLID). This has severe implications, which breaks the purpose of having the document store abstraction in the first place, because the implementors now have to make their client code work differently for different underlying datastore. And, honestly, if people miss to read this code comment, and expect the same client-code to work perfectly fine in the other world (which is highly likely), it can lead to subtle bugs, hard to debug and reproduce.

The older methods (like update) have in fact violated this leading to the introduction of newer consistent methods. Perhaps, the behaviour can be restrictive (to at least the known databases which has the most restrictive atomicity support), but has to be consistent.

suddendust: @suresh-prakash Thanks for the detailed comment, totally agree with you. I feel batch-level atomicity could be controlled by adding them in the `UpdateOptions`. Or, this method's contract can say that `batch-level atomicity is not guaranteed, while per-key update atomicity has to be guaranteed`. This is the least-restrictive contract that almost any implementation can implement. If any implementation wants to provide batch-level atomicity guarantees, that's even better. 

Wdyt?

suresh-prakash: > batch-level atomicity is not guaranteed, while per-key update atomicity has to be guaranteed

Yes. This makes total sense. 

> If any implementation wants to provide batch-level atomicity guarantees, that's even better.

Better definitely in terms of performance. But, that can create functional digression, especially when one of the multiple updates is invalid. In one case (supporting per-key atomicity), documents before that key would have been updated. While, in another case (supporting batch-level atomicity), the entire update would fail.

If we are using `document-store`, all implementors have to be consistent as much as possible in terms of functionality. Performance would (and can) differ based on the underlying database used, indexes created, etc.

From this standpoint, I think, it would make sense to implement Postgres also perform key-level atomic updates (at least by default). If performance is critical the client can opt-in as you mentioned via. `UpdateOptions` with the implications explicitly known.

Basically, the design goal here is that the caller/client should be aware of the implications even when they missed to read out these comments and hence prefer cross-database consistency over performance (since the performance differences are evident and inevitable).

suddendust: Makes sense, I have updated the doc.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
```diff
@@ -896,6 +907,109 @@ public CloseableIterator<Document> bulkUpdate(
...
+
+    Preconditions.checkArgument(updateOptions != null, "UpdateOptions cannot be NULL");
+
+    String tableName = tableIdentifier.getTableName();
+    String quotedPkColumn = PostgresUtils.wrapFieldNamesWithDoubleQuotes(getPKForTable(tableName));
+
+    Set<Key> updatedKeys = new HashSet<>();
+
+    try (Connection connection = client.getPooledConnection()) {
+      for (Map.Entry<Key, Collection<SubDocumentUpdate>> entry : updates.entrySet()) {
```
suresh-prakash, line 927: Based on our conversation @ https://github.com/hypertrace/document-store/pull/288#discussion_r3072575124, seems like this is making the whole operations (batch-level) atomic (having a different variant than the Mongo implementation to come in the future).

Could you please check?

suddendust: Is it? For batch-level transactions, we would acquire a `client.getTransactionalConnection()`. Might be wrong here but afai understand it's atomic for each key, not the entire batch.

suresh-prakash: My bad. I thought you are doing transactions here. You are right.

suresh-prakash (approved):

Issue #290, opened 2026-04-27
Fix Query Shape for FlatPostgresCollection#upsert and createOrReplace for Query Plan Caching


suddendust:
We observed that the query shape is different and because of this, the query plan isn't cached. This needs to be fixed.

PR #299, opened 2026-04-28
[290] Fix query shape for FlatPostgresCollection upsert and createOrReplace

suddendust:
## Description
Ref: https://github.com/hypertrace/document-store/issues/290

## Stable SQL shape for flat collection upsert/createOrReplace

Previously, the SQL generated for `upsert` and `createOrReplace` operations varied based on the document's fields. This prevented PgJDBC from promoting queries to server-side prepared statements, since each unique SQL string is treated as a new statement.

Now, all four write methods (`upsert`, `bulkUpsert`, `createOrReplace`, `bulkCreateOrReplace`) always use all table columns in the generated SQL, producing a stable shape that depends only on the table schema.

### Upsert (merge semantics)

**Before:**
```sql
INSERT INTO "myTable" ("id", "price")
VALUES (?, ?)
ON CONFLICT ("id") DO UPDATE SET "price" = COALESCE(EXCLUDED."price", "myTable"."price")
```

**After:**
```sql
INSERT INTO "myTable" ("id", "price", "item", "quantity")
VALUES (?, ?, ?, ?)
ON CONFLICT ("id") DO UPDATE SET
  "price" = COALESCE(EXCLUDED."price", "myTable"."price"),
  "item" = COALESCE(EXCLUDED."item", "myTable"."item"),
  "quantity" = COALESCE(EXCLUDED."quantity", "myTable"."quantity")
```

Missing columns are bound as `NULL`; `COALESCE` preserves existing values.

### CreateOrReplace (full-replace semantics)

**Before:**
```sql
INSERT INTO "myTable" ("id", "price")
VALUES (?, ?)
ON CONFLICT ("id") DO UPDATE SET
  "price" = EXCLUDED."price",
  "item" = DEFAULT,
  "quantity" = DEFAULT
```

**After:**
```sql
INSERT INTO "myTable" ("id", "price", "item", "quantity")
VALUES (?, ?, ?, ?)
ON CONFLICT ("id") DO UPDATE SET
  "price" = EXCLUDED."price",
  "item" = EXCLUDED."item",
  "quantity" = EXCLUDED."quantity"
```

Missing columns are bound as `NULL`; `EXCLUDED.col` overwrites with `NULL` (full-replace).

Essentially, both `upsert` and `createOrReplace` bind `NULL` for missing columns. What changes is how the `SET` clause handles that `NULL`:

`Upsert`: `SET col = COALESCE(EXCLUDED.col, table.col)` → `NULL` is ignored, existing value kept
`CreateOrReplace`: `SET col = EXCLUDED.col` → `NULL` overwrites the existing value

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
@@ -392,7 +392,7 @@ public class FlatPostgresCollection extends PostgresCollection {
       // Build the bulk upsert SQL with all columns
       // Use COALESCE to preserve existing values when a document doesn't have a column
       List<String> columnList = new ArrayList<>(allColumns);
-      String sql = buildMergeUpsertSql(columnList, quotedPkColumn, false, true);
+      String sql = buildMergeUpsertSql(columnList, quotedPkColumn, false);
       LOGGER.debug("Bulk upsert SQL: {}", sql);
 
       try (Connection conn = client.getPooledConnection();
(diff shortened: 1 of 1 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
```diff
@@ -1734,6 +1678,13 @@ private PostgresDataType getPrimaryKeyType(String tableName, String pkColumn) {
         .orElse(PostgresDataType.TEXT);
   }
 
+  private List<String> getAllQuotedColumns(String tableName) {
+    return schemaRegistry.getSchema(tableName).values().stream()
+        .map(PostgresColumnMetadata::getName)
+        .map(PostgresUtils::wrapFieldNamesWithDoubleQuotes)
+        .collect(Collectors.toList());
```
puneet-traceable, line 1685: can we avoid this computation again and again?

suddendust: Created a separate issue for this: https://github.com/hypertrace/document-store/issues/301

Would like to profile this once across multiple upsert calls in the load test and based off that, will come up with a plan.

puneet-traceable (approved):

PR #308, opened 2026-05-20
Flatcollection search queryconverter bugfix

suddendust:
### Problem

The legacy `Collection.search()` path goes through `LegacyFilterToQueryFilterTransformer` (for filters) and `LegacyQueryToV2QueryTransformer` (for selections / orderBy) to convert the v1 `Filter`/`Query` API into the v2 `Query` API before handing off to the Postgres parsers. Both transformers were dropping the column's type information when they constructed the v2 `IdentifierExpression`, with two visible consequences:

1. **Array columns produced invalid SQL.** A filter like `Filter(EQ, "tags", "hygiene")` against a `TEXT[]` column was rendered as `"tags" = ?` with a text RHS — Postgres rejects this at execution time because `text = text[]` has no operator. The same problem occurred for `INTEGER[]`, `DOUBLE PRECISION[]`, `BOOLEAN[]`, and any `IN`/`NOT_IN` variants on array columns.

2. **JSONB column resolution was ad-hoc.** Both transformers separately walked the schema looking for a JSONB prefix and built `JsonIdentifierExpression` directly, with no shared codepath for the "field resolves to the JSONB column itself" case (which is forbidden — `JsonIdentifierExpression` requires a non-empty `jsonPath`).

Net effect: filters on array columns failed at runtime, and the same logical query that worked through the v2 `find()` API blew up through the legacy `search()` API.

### Solution

Introduced **`IdentifierExpressionFactory`** as a single resolution point that maps a `(fieldName, ColumnMetadata)` pair to the correct `IdentifierExpression` subtype using the column's `canonicalType` and `isArray()` signals from the schema registry.

```
ColumnMetadata.canonicalType + isArray
        ↓
IdentifierExpressionFactory.createIdentifierFromColumn
        ↓
- ArrayIdentifierExpression.of{Strings,Ints,Longs,Floats,Doubles,Booleans,TimestampsTz,Dates}  (array columns)
- IdentifierExpression.of{String,Int,Long,Float,Double,Boolean,TimestampTz,Date}              (scalar columns)
- IdentifierExpression.of(name)                                                               (JSONB column itself, or no type info)
```

Three cases the factory has to handle correctly:

| Case | Returned type | Why |
|---|---|---|
| Typed scalar column (`STRING`, `INTEGER`, …) | Typed `IdentifierExpression.ofX(name)` | Parser needs the element type to emit a correctly-cast comparison. |
| Typed array column (any of the above, `isArray() == true`) | Typed `ArrayIdentifierExpression.ofXs(name)` | Parser routes to array-aware SQL (`= ANY(...)`, `&&`, `@>`). |
| Whole JSONB column referenced by name (no nested path) | Untyped `IdentifierExpression.of(name)` | `JsonIdentifierExpression` requires a non-empty `jsonPath` — it represents a path *inside* a JSONB column, not the column itself. |
| Unknown / no schema info | Untyped `IdentifierExpression.of(name)` | Fall through to runtime resolution, behavior matches pre-fix. |

Both `LegacyFilterToQueryFilterTransformer` and `LegacyQueryToV2QueryTransformer` now route their direct-column lookups through this factory. The JSONB-prefix walk (for nested paths like `props.brand`) remains in each transformer because the value-type inference (which becomes `JsonFieldType`) is op-specific:

- `LegacyFilterToQueryFilterTransformer.inferJsonFieldType(value)` looks at the filter value's runtime type — `String → STRING`, `Number → NUMBER`, `Boolean → BOOLEAN`. **Note:** there is no `STRING_ARRAY` inference path today; array-membership semantics on nested JSONB arrays are not reachable from the legacy filter API. Documented in tests.
- `LegacyQueryToV2QueryTransformer` (selections / orderBy) defaults to `JsonFieldType.STRING` since there's no value to infer from.

### DIFF
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -6296,6 +6296,126 @@ public class DocStoreQueryV1Test {
             }
           });
     }
+
+    // The tests below exercise the legacy search() path against Postgres top-level array columns.
+    // Before the IdentifierExpressionFactory fix, the legacy v1 -> v2 transformer dropped column
+    // type information, so filters on text[] / integer[] / double precision[] columns rendered
+    // SQL that Postgres rejected at execution time (e.g. "tags IN (?)" with a text RHS against a
+    // text[] column). With the fix, the transformer now emits an ArrayIdentifierExpression with
+    // the correct element type, so the Postgres parsers generate array-aware SQL (&&, @>, =).
+    @ParameterizedTest
+    @ArgumentsSource(PostgresProvider.class)
+    void testSearchWithInOnTextArrayColumn(String dataStoreName) {
+      Collection flatCollection = getFlatCollection(dataStoreName);
+
+      org.hypertrace.core.documentstore.Query legacyQuery =
+          new org.hypertrace.core.documentstore.Query()
+              .withFilter(
+                  new org.hypertrace.core.documentstore.Filter(
+                      org.hypertrace.core.documentstore.Filter.Op.IN,
+                      "tags",
+                      List.of("hygiene", "grooming")));
+
+      Iterator<Document> results = flatCollection.search(legacyQuery);
+      int count = 0;
+      while (results.hasNext()) {
+        results.next();
+        count++;
+      }
+      // ids 1, 5, 8 (hygiene) + ids 6, 7 (grooming) = 5 docs whose tags array overlaps the filter
+      assertEquals(5, count);
+    }
+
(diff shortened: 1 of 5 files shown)


puneet-traceable (approved):

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -5975,326 +5975,409 @@ private static Collection getFlatCollection(String dataStoreName) {
   @Nested
   class FlatCollectionLegacySearchMethod {
```
suddendust, line 5976: The strategy here is to run the same v1 and v2 query again the collection and both should return the same results.

PR #312, opened 2026-06-19
Changes to enhance existing length op

sarthak77:
## Summary

Enhances the `LENGTH` function so that taking the length of an **absent, null, or empty** array field returns `0` instead of erroring — consistently across MongoDB and Postgres, and whether the field is sorted on, selected, or filtered.

Previously `LENGTH` on a missing/empty array could throw (`$size` on null in Mongo; `array_length(...)` type errors in Postgres), making queries that sort by computed list size fail.

## Behavior

`LENGTH(<arrayField>)` now returns:

| Case | Result |
| --- | --- |
| Field missing / absent | `0` |
| Field present but JSON `null` | `0` |
| Empty array (`[]` / `{}`) | `0` |
| Non-empty array | element count |

Works in `ORDER BY`, selections, and aggregation filters.

## Changes

**MongoDB** (`MongoFunctionExpressionParser`)
- Wrap the `$size` operand in `$ifNull: [<expr>, []]` so a missing/null field defaults to an empty array → `0`.

**Postgres** — length SQL is chosen by the field's storage layout, with the decision owned by the column transformer (not the function visitor):
- Added `PostgresColTransformer.buildArrayLengthExpression(...)`.
  - `FlatPostgresFieldTransformer`: native array column (`tags TEXT[]`) → `COALESCE(ARRAY_LENGTH(...), 0)`; array nested in a JSONB column (`props.colors`) → `jsonb_array_length(...)`.
  - `NestedPostgresColTransformer`: always `jsonb_array_length(...)`.
- JSONB lengths use a `CASE WHEN jsonb_typeof(...) = 'array' THEN ... ELSE '[]'::jsonb END` guard so SQL `NULL`, JSON `null`, and non-array values all resolve to `0`.
- Shared templates extracted into `PostgresUtils.prepareArrayLength(...)` / `prepareJsonbArrayLength(...)`.
- `LENGTH` over an aggregation alias (`ARRAY_AGG`) keeps `ARRAY_LENGTH`, now wrapped in `COALESCE(..., 0)`.

## Tests

- `testSortByListSizeWithMissingField` (Mongo + Postgres) — sort by `LENGTH` with a missing field; asserts `0` and sorts first.
- `testSortByJsonbArrayLengthInFlatCollection` (Postgres) — `LENGTH` on an array nested in a JSONB column via typed `JsonIdentifierExpression`; covers multi-element, empty, and NULL.
- `testSortByTopLevelArrayLengthInFlatCollection` (Postgres) — `LENGTH` on a native `TEXT[]` column; covers NULL and empty → `0`.
- Updated expected Mongo pipeline fixtures and `PostgresQueryParserTest` SQL assertions.

### DIFF
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -928,44 +928,47 @@ public class DocStoreQueryV1Test {
     datastore.createCollection(collectionName, null);
     Collection collection = datastore.getCollection(collectionName);
 
-    collection.upsert(
-        new SingleValueKey(TENANT_ID, "three"),
-        new JSONDocument("{\"item\":\"three\",\"tags\":[\"a\",\"b\",\"c\"]}"));
-    collection.upsert(
-        new SingleValueKey(TENANT_ID, "one"),
-        new JSONDocument("{\"item\":\"one\",\"tags\":[\"x\"]}"));
-    // Document intentionally missing the "tags" field; LENGTH must resolve to 0 instead of failing
-    collection.upsert(
-        new SingleValueKey(TENANT_ID, "none"), new JSONDocument("{\"item\":\"none\"}"));
+    try {
+      collection.upsert(
+          new SingleValueKey(TENANT_ID, "three"),
+          new JSONDocument("{\"item\":\"three\",\"tags\":[\"a\",\"b\",\"c\"]}"));
+      collection.upsert(
+          new SingleValueKey(TENANT_ID, "one"),
+          new JSONDocument("{\"item\":\"one\",\"tags\":[\"x\"]}"));
+      // Document intentionally missing the "tags" field; LENGTH must resolve to 0 instead of
+      // failing
+      collection.upsert(
+          new SingleValueKey(TENANT_ID, "none"), new JSONDocument("{\"item\":\"none\"}"));
 
-    Query query =
-        Query.builder()
-            .addSelection(IdentifierExpression.of("item"))
-            .addSelection(
-                FunctionExpression.builder()
-                    .operator(LENGTH)
-                    .operand(IdentifierExpression.of("tags"))
-                    .build(),
-                "tag_count")
-            .addSort(IdentifierExpression.of("tag_count"), ASC)
-            .build();
+      Query query =
+          Query.builder()
+              .addSelection(IdentifierExpression.of("item"))
+              .addSelection(
+                  FunctionExpression.builder()
+                      .operator(LENGTH)
+                      .operand(IdentifierExpression.of("tags"))
+                      .build(),
+                  "tag_count")
+              .addSort(IdentifierExpression.of("tag_count"), ASC)
+              .build();
 
-    Iterator<Document> resultDocs = collection.aggregate(query);
-    List<Map<String, Object>> results = new ArrayList<>();
-    while (resultDocs.hasNext()) {
-      results.add(Utils.convertDocumentToMap(resultDocs.next()));
+      Iterator<Document> resultDocs = collection.aggregate(query);
+      List<Map<String, Object>> results = new ArrayList<>();
+      while (resultDocs.hasNext()) {
+        results.add(Utils.convertDocumentToMap(resultDocs.next()));
+      }
+
+      assertEquals(3, results.size());
+      // Document without the "tags" field counts as 0 and sorts first in ascending order
+      assertEquals("none", results.get(0).get("item"));
+      assertEquals(0, ((Number) results.get(0).get("tag_count")).intValue());
+      assertEquals("one", results.get(1).get("item"));
+      assertEquals(1, ((Number) results.get(1).get("tag_count")).intValue());
+      assertEquals("three", results.get(2).get("item"));
+      assertEquals(3, ((Number) results.get(2).get("tag_count")).intValue());
+    } finally {
+      datastore.deleteCollection(collectionName);
     }
-
-    assertEquals(3, results.size());
-    // Document without the "tags" field counts as 0 and sorts first in ascending order
-    assertEquals("none", results.get(0).get("item"));
-    assertEquals(0, ((Number) results.get(0).get("tag_count")).intValue());
-    assertEquals("one", results.get(1).get("item"));
-    assertEquals(1, ((Number) results.get(1).get("tag_count")).intValue());
-    assertEquals("three", results.get(2).get("item"));
-    assertEquals(3, ((Number) results.get(2).get("tag_count")).intValue());
-
-    datastore.deleteCollection(collectionName);
   }
 
   @ParameterizedTest
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -985,7 +985,8 @@ public class DocStoreQueryV1Test {
             .addSelection(
                 FunctionExpression.builder()
                     .operator(LENGTH)
-                    .operand(JsonIdentifierExpression.of("props", "colors"))
+                    .operand(
+                        JsonIdentifierExpression.of("props", JsonFieldType.STRING_ARRAY, "colors"))
                     .build(),
                 "color_count")
             .addSort(IdentifierExpression.of("color_count"), DESC)
@@ -1009,6 +1010,45 @@ public class DocStoreQueryV1Test {
     assertEquals(1, ((Number) results.get(2).get("color_count")).intValue());
     // Remaining have 0 (empty array or NULL props)
     assertEquals(0, ((Number) results.get(3).get("color_count")).intValue());
+  }
+
+  @ParameterizedTest
+  @ArgumentsSource(PostgresProvider.class)
+  public void testSortByTopLevelArrayLengthInFlatCollection(String dataStoreName)
+      throws IOException {
+    Collection flatCollection = getFlatCollection(dataStoreName);
+
+    // "tags" is a native PG array column (TEXT[]) — a top-level field, not nested in a JSONB
+    // column.
+    // Test data: id=3 has 4 tags; id=1,2,4,5,6,7,8 have 3 tags; id=10 is empty {}; id=9 is NULL.
+    Query query =
+        Query.builder()
+            .addSelection(IdentifierExpression.of("id"))
+            .addSelection(
+                FunctionExpression.builder()
+                    .operator(LENGTH)
+                    .operand(IdentifierExpression.of("tags"))
+                    .build(),
+                "tag_count")
+            .addSort(IdentifierExpression.of("tag_count"), ASC)
+            .addSort(IdentifierExpression.of("id"), ASC)
+            .build();
+
+    Iterator<Document> resultDocs = flatCollection.aggregate(query);
+    List<Map<String, Object>> results = new ArrayList<>();
+    while (resultDocs.hasNext()) {
+      results.add(Utils.convertDocumentToMap(resultDocs.next()));
+    }
+
+    assertEquals(10, results.size());
+    // NULL (id=9) and empty {} (id=10) both resolve to 0 and sort first (id ASC: "10" < "9").
+    assertEquals("10", results.get(0).get("id"));
+    assertEquals(0, ((Number) results.get(0).get("tag_count")).intValue());
+    assertEquals("9", results.get(1).get("id"));
+    assertEquals(0, ((Number) results.get(1).get("tag_count")).intValue());
+    // id=3 has the most tags (4) and sorts last.
+    assertEquals("3", results.get(9).get("id"));
+    assertEquals(4, ((Number) results.get(9).get("tag_count")).intValue());
   }
 
   @ParameterizedTest
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -920,6 +920,55 @@ public class DocStoreQueryV1Test {
   }
 
   @ParameterizedTest
+  @ArgumentsSource(MongoProvider.class)
+  public void testSortByListSizeWithMissingField(String dataStoreName) throws IOException {
+    Datastore datastore = datastoreMap.get(dataStoreName);
+    String collectionName = "list_size_sort_collection";
+    datastore.deleteCollection(collectionName);
+    datastore.createCollection(collectionName, null);
+    Collection collection = datastore.getCollection(collectionName);
+
+    collection.upsert(
+        new SingleValueKey(TENANT_ID, "three"),
+        new JSONDocument("{\"item\":\"three\",\"tags\":[\"a\",\"b\",\"c\"]}"));
+    collection.upsert(
+        new SingleValueKey(TENANT_ID, "one"),
+        new JSONDocument("{\"item\":\"one\",\"tags\":[\"x\"]}"));
+    // Document intentionally missing the "tags" field; LENGTH must resolve to 0 instead of failing
+    collection.upsert(
+        new SingleValueKey(TENANT_ID, "none"), new JSONDocument("{\"item\":\"none\"}"));
+
+    Query query =
+        Query.builder()
+            .addSelection(IdentifierExpression.of("item"))
+            .addSelection(
+                FunctionExpression.builder()
+                    .operator(LENGTH)
+                    .operand(IdentifierExpression.of("tags"))
+                    .build(),
+                "tag_count")
+            .addSort(IdentifierExpression.of("tag_count"), ASC)
+            .build();
+
+    Iterator<Document> resultDocs = collection.aggregate(query);
+    List<Map<String, Object>> results = new ArrayList<>();
+    while (resultDocs.hasNext()) {
+      results.add(Utils.convertDocumentToMap(resultDocs.next()));
+    }
+
+    assertEquals(3, results.size());
+    // Document without the "tags" field counts as 0 and sorts first in ascending order
+    assertEquals("none", results.get(0).get("item"));
+    assertEquals(0, ((Number) results.get(0).get("tag_count")).intValue());
+    assertEquals("one", results.get(1).get("item"));
+    assertEquals(1, ((Number) results.get(1).get("tag_count")).intValue());
+    assertEquals("three", results.get(2).get("item"));
+    assertEquals(3, ((Number) results.get(2).get("tag_count")).intValue());
+
+    datastore.deleteCollection(collectionName);
+  }
+
+  @ParameterizedTest
   @ArgumentsSource(AllProvider.class)
   public void testAggregateWithFunctionalLeftHandSideFilter(final String dataStoreName)
       throws IOException {
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
@@ -920,7 +920,7 @@ public class DocStoreQueryV1Test {
   }
 
   @ParameterizedTest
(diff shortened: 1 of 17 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +84,20 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
+        identifier.map(v -> getPostgresQueryParser().getPgSelections().get(v));
+    if (resolvedSelection.isPresent()) {
+      return String.format(
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", resolvedSelection.get(), ARRAY_DIMENSION);
+    }
+    PostgresFieldIdentifierExpressionVisitor jsonbVisitor =
+        new PostgresFieldIdentifierExpressionVisitor(getPostgresQueryParser());
+    String parsedExpression = operand.accept(jsonbVisitor);
+    return String.format("COALESCE( jsonb_array_length( %s ), 0 )", parsedExpression);
+  }
```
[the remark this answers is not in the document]
sarthak77: Addressed in b7b7439. The code now checks `getDocumentType()` on the `pgColTransformer`:
- **FLAT** collections (native PG arrays like `TEXT[]`): uses `ARRAY_LENGTH(column, 1)`
- **NESTED** collections (jsonb document fields): uses `jsonb_array_length(document->'field')`

Both are wrapped with `COALESCE(..., 0)` so NULL/missing fields return 0.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -919,6 +919,55 @@ public void testAggregateWithMultipleGroupingLevels(String dataStoreName) throws
     testCountApi(dataStoreName, query, "query/multi_level_grouping_response.json");
   }
 
+  @ParameterizedTest
+  @ArgumentsSource(AllProvider.class)
+  public void testSortByListSizeWithMissingField(String dataStoreName) throws IOException {
+    Datastore datastore = datastoreMap.get(dataStoreName);
+    String collectionName = "list_size_sort_collection";
```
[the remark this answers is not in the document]
sarthak77: The integration test runs against nested collections (the standard document store pattern) which exercises the `jsonb_array_length` path. The flat collection path (`ARRAY_LENGTH`) is covered by the existing unit tests in `PostgresQueryParserTest` which validate SQL generation for LENGTH on aggregation aliases. The flat collection LENGTH on a direct column field follows the same `ARRAY_LENGTH` codepath and is validated by the existing tests that assert correct SQL output.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -919,6 +919,55 @@ public void testAggregateWithMultipleGroupingLevels(String dataStoreName) throws
...
+    datastore.deleteCollection(collectionName);
+    datastore.createCollection(collectionName, null);
+    Collection collection = datastore.getCollection(collectionName);
```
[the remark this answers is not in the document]
sarthak77: Good catch. Wrapped the test body in try-finally in be9514a so `deleteCollection` runs regardless of assertion failures.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,24 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", resolvedSelection.get(), ARRAY_DIMENSION);
+    }
+    PostgresFieldIdentifierExpressionVisitor fieldVisitor =
+        new PostgresFieldIdentifierExpressionVisitor(getPostgresQueryParser());
+    String parsedExpression = operand.accept(fieldVisitor);
+    if (getPostgresQueryParser().getPgColTransformer().getDocumentType() == DocumentType.FLAT) {
+      return String.format(
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", parsedExpression, ARRAY_DIMENSION);
+    }
+    return String.format("COALESCE( jsonb_array_length( %s ), 0 )", parsedExpression);
```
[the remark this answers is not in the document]
sarthak77: Fixed in bc24f4c. Replaced `COALESCE(jsonb_array_length(...), 0)` with the `CASE WHEN jsonb_typeof(...) = 'array' THEN ... ELSE '[]'::jsonb END` pattern already used elsewhere in the codebase (e.g. `PostgresFromTypeExpressionVisitor`, `PostgresFilterTypeExpressionVisitor`). This handles SQL NULL (missing key), JSON null, and non-array types — all return 0.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,27 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
   }
 
+  private String buildLengthExpression(final SelectTypeExpression operand) {
+    Optional<String> identifier = Optional.ofNullable(operand.accept(identifierExpressionVisitor));
+    Optional<String> resolvedSelection =
+        identifier.map(v -> getPostgresQueryParser().getPgSelections().get(v));
+    if (resolvedSelection.isPresent()) {
+      return String.format(
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", resolvedSelection.get(), ARRAY_DIMENSION);
+    }
```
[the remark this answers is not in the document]
sarthak77: In practice, the `pgSelections` path is only reached when LENGTH operates on an alias defined in a prior selection — and those aliases always come from aggregation expressions (`ARRAY_AGG`, `DISTINCT_ARRAY`) that produce native PG arrays. There's no existing usage pattern where a raw jsonb field is aliased and then LENGTH is applied to that alias. The direct field access case (without an alias) is handled by the branch below which uses `jsonb_typeof` for nested collections.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -49,10 +50,11 @@ public String visit(final FunctionExpression expression) {
     }
 
     if (numArgs == 1) {
+      if (expression.getOperator().equals(FunctionOperator.LENGTH)) {
```
puneet-traceable: did we have same problem here with length function?

sarthak77: Tests were failing for postgres without the changes.

puneet-traceable: why are the tests failing when there is no change to postgres path?

sarthak77: The new integration test `testSortByListSizeWithMissingField` runs on both Mongo and Postgres (via `@ArgumentsSource(AllProvider.class)`). The test applies LENGTH on a raw jsonb field (`tags`) where one document is missing that field entirely.

The original Postgres LENGTH implementation used `ARRAY_LENGTH(parsedExpression, 1)` — but `parsedExpression` here resolved via `getParsedExpression()` which uses `PostgresDataAccessorIdentifierExpressionVisitor`. That visitor casts the jsonb field to NUMERIC (via `->>` + `CAST`), resulting in `ARRAY_LENGTH(CAST(document->>'tags' AS NUMERIC), 1)` — which fails with `function array_length(numeric, integer) does not exist`.

Even before this PR, calling LENGTH on a raw jsonb field in Postgres would have failed the same way — it just wasn't tested. The fix distinguishes between native PG arrays (from aggregations) and raw jsonb fields, using `jsonb_array_length` with a `jsonb_typeof` guard for the latter.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,27 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", resolvedSelection.get(), ARRAY_DIMENSION);
+    }
+    PostgresFieldIdentifierExpressionVisitor fieldVisitor =
+        new PostgresFieldIdentifierExpressionVisitor(getPostgresQueryParser());
+    String parsedExpression = operand.accept(fieldVisitor);
+    if (getPostgresQueryParser().getPgColTransformer().getDocumentType() == DocumentType.FLAT) {
+      return String.format(
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", parsedExpression, ARRAY_DIMENSION);
+    }
+    return String.format(
```
puneet-traceable: @sarthak77 Can you please attempt to test a case where the field to be use with length function is inside a column which is a jsonb column. Flat collections can have a column which is jsonb and custom attributes will be placed there. So, from my understanding the check above `getPostgresQueryParser().getPgColTransformer().getDocumentType() == DocumentType.FLAT` to determine whether to use `jsonb_array_length` is not right.

sarthak77: Sure. Let me add that. I'm not able to run tests locally so might push few test commits.

sarthak77: You were right — the `DocumentType.FLAT` check was incorrect. Added an integration test `testSortByJsonbArrayLengthInFlatCollection` that uses `JsonIdentifierExpression.of("props", "colors")` on the flat collection. It failed with `array_length(jsonb, integer) does not exist` as expected.

Fixed in e75d045 — now checks `FieldToPgColumn.getTransformedField()` instead:
- `transformedField == null` → direct column (native PG array) → `ARRAY_LENGTH`
- `transformedField != null` → jsonb path inside a column → `jsonb_array_length`

CI test job passes now.

puneet-traceable: I think we should fix it here: https://github.com/hypertrace/document-store/blob/a02a27993152909a27fc2563d8d486815ecadac3/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FlatPostgresFieldTransformer.java

Current impl leaks storage‑layout concerns into the function visitor: the visitor now calls transformField(...) and peeks at FieldToPgColumn.transformedField, which is exactly what the transformer was introduced to hide.

sarthak77: Agreed. Moved the logic into the transformer in 5fb58fa. Added `buildArrayLengthExpression(FieldToPgColumn)` to the `PostgresColTransformer` interface:
- `FlatPostgresFieldTransformer` → `ARRAY_LENGTH` for direct columns, `jsonb_array_length` for jsonb paths
- `NestedPostgresColTransformer` → always `jsonb_array_length`

The visitor now just delegates via `transformField(...)` + `buildArrayLengthExpression(...)` and no longer inspects `FieldToPgColumn.transformedField`, so storage-layout concerns stay in the transformer.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,29 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
+      return String.format(
+          "COALESCE( ARRAY_LENGTH( %s, %s ), 0 )", resolvedSelection.get(), ARRAY_DIMENSION);
+    }
+    if (operand instanceof IdentifierExpression) {
+      PostgresQueryParser parser = getPostgresQueryParser();
+      return parser
+          .getPgColTransformer()
+          .buildArrayLengthExpression(parser.transformField((IdentifierExpression) operand));
+    }
+    PostgresFieldIdentifierExpressionVisitor fieldVisitor =
```
puneet-traceable: Which case does line 102-108 hit? did you check from tests?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FlatPostgresFieldTransformer.java
```diff
@@ -103,6 +103,18 @@ public String buildFieldAccessorWithoutCast(FieldToPgColumn fieldToPgColumn) {
...
   }
 
+  @Override
+  public String buildArrayLengthExpression(FieldToPgColumn fieldToPgColumn) {
+    String fieldAccessor = buildFieldAccessorWithoutCast(fieldToPgColumn);
+    if (fieldToPgColumn.getTransformedField() == null) {
+      return String.format("COALESCE( ARRAY_LENGTH( %s, 1 ), 0 )", fieldAccessor);
+    }
+    return String.format(
+        "jsonb_array_length( CASE WHEN jsonb_typeof( %s ) = 'array' THEN %s"
```
puneet-traceable: This template is at multiple places, can it move to some utility if there is no better way.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,29 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
         String.format("Query operation:%s not supported", operator));
   }
 
+  private String buildLengthExpression(final SelectTypeExpression operand) {
```
puneet-traceable: the interface method's intent is arrayLength while this function just says length, should be consistent.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
```diff
@@ -83,6 +85,29 @@ private Collector getCollectorForFunctionOperator(FunctionOperator operator) {
...
+    Optional<String> identifier = Optional.ofNullable(operand.accept(identifierExpressionVisitor));
```
puneet-traceable: I do not think this optional conversion adds any value here. Simpler null check instead of `resolvedSelection.isPresent()` should be fine.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
sarthak77: Good catch — that fallback branch was dead. LENGTH operands are always either a selection alias or a field identifier (both `IdentifierExpression`), never a non-identifier expression. Removed it in 08650ab; the method now handles only the alias and field-identifier cases.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/transformer/FlatPostgresFieldTransformer.java
sarthak77: Moved both templates into `PostgresUtils` in 08650ab — `prepareArrayLength(...)` and `prepareJsonbArrayLength(...)`. Both transformers and the visitor now call those instead of inlining the format strings.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
sarthak77: Renamed the private method to `buildArrayLengthExpression` in 08650ab to match the interface method's intent.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFunctionExpressionVisitor.java
sarthak77: Replaced the Optional plumbing with a plain null check in 08650ab.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -919,6 +919,98 @@ public void testAggregateWithMultipleGroupingLevels(String dataStoreName) throws
...
+    // props.colors is a jsonb array inside a JSONB column in the flat collection.
+    // Test data: id=1 ["Blue","Green"], id=3 ["Black"], id=5 ["Orange","Blue"],
+    // id=7 [] (empty), id=2,4,6,8,9,10 props=NULL.
+    Query query =
+        Query.builder()
+            .addSelection(IdentifierExpression.of("id"))
+            .addSelection(
+                FunctionExpression.builder()
+                    .operator(LENGTH)
+                    .operand(JsonIdentifierExpression.of("props", "colors"))
```
suddendust: Can you try this test with `JsonIdentifierExpression.of("props", JsonFieldType.STRING_ARRAY, "colors")`?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -919,6 +919,98 @@ public void testAggregateWithMultipleGroupingLevels(String dataStoreName) throws
...
+      assertEquals("three", results.get(2).get("item"));
+      assertEquals(3, ((Number) results.get(2).get("tag_count")).intValue());
+    } finally {
+      datastore.deleteCollection(collectionName);
+    }
+  }
+
+  @ParameterizedTest
+  @ArgumentsSource(PostgresProvider.class)
+  public void testSortByJsonbArrayLengthInFlatCollection(String dataStoreName) throws IOException {
```
suddendust: Can you also add a test case for sorting by top-level array length?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
sarthak77: Done in a77362a — the test now uses `JsonIdentifierExpression.of("props", JsonFieldType.STRING_ARRAY, "colors")`. The LENGTH path resolves the field via the transformer (column + json path), so it produces the same `jsonb_array_length(...)` SQL regardless of the declared `JsonFieldType`; the explicit type just makes the intent clearer.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
sarthak77: Added `testSortByTopLevelArrayLengthInFlatCollection` in a77362a. It sorts by LENGTH of the native `tags TEXT[]` column (top-level, not nested), asserting NULL (id=9) and empty `{}` (id=10) both resolve to 0 and the 4-element row (id=3) sorts last. This exercises the `ARRAY_LENGTH` branch end-to-end, complementing the JSONB `jsonb_array_length` case.

suddendust (approved):
LGTM

PR #320, opened 2026-07-21
[FlatPostgresCollection] Sort keys before preparedStatement.addBatch() to prevent deadlocks

suddendust:
# Sort keys to prevent deadlocks in batched writes on FlatPostgresCollection

## Summary

 Enforces a canonical, string-sorted key order across all JDBC-batched, key-keyed write operations in FlatPostgresCollection. Postgres acquires per-row locks in batch-entry order for `INSERT ... ON CONFLICT DO UPDATE` and `UPDATE ... WHERE key = ?`, so two concurrent batches that touch overlapping key sets in opposite orders can form a wait-for cycle and one side gets aborted with ERROR: deadlock detected (SQLState `40P01`).

  Sorting inside the store — the last hop before `addBatch(...)` — makes the invariant local to the collection and independent of caller iteration order, so no downstream service has to remember to sort.

  Symptom this fixes

  Callers were seeing production log lines like:

```sql
Failed to execute batch update. SQL: UPDATE "myTable" SET "attributes.col1" = ?, "col2" = ? WHERE "pkCol" = ?
Error: Batch entry 8 UPDATE "myTable" ... was aborted:
ERROR: deadlock detected.
```

The offending SQL is produced only by the Map-based `bulkUpdate` on this class, and the deadlock reproduces trivially with two concurrent callers that iterate overlapping keys in opposite orders. The same class of bug also exists on `bulkUpsert` / `bulkCreateOrReplace`, which batch `INSERT ... ON CONFLICT DO UPDATE — the DO UPDATE` branch takes a row-exclusive lock and follows the same batch-entry ordering.

### Testing
Didn't add any additional tests - The existing integration tests should be sufficient.

### Checklist:

### DIFF
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
@@ -1032,6 +1032,7 @@ public class FlatPostgresCollection extends PostgresCollection {
     for (int i = 0; i < order.length; i++) {
       order[i] = i;
     }
+    // Sort keys to ensure consistent order to avoid deadlocks
     Arrays.sort(order, Comparator.comparing(i -> keys.get(i).toString()));
 
     List<String> setFragments = new ArrayList<>(keyGroup.getSetFragments());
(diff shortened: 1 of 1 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/FlatPostgresCollection.java
```diff
@@ -1022,6 +1029,12 @@ private int executeBatchUpdate(
     List<Key> keys = keyGroup.getKeys();
     List<List<Object>> allKeyParams = keyGroup.getKeyParams();
 
+    Integer[] order = new Integer[keys.size()];
```
suddendust, line 1031: This array contains the order of the keys - We add to the batch according to this order.

suresh-prakash (approved):

PR #322, opened 2026-08-24
Return false from bulkUpsert when write result is incomplete

Bhuvan506:
## Summary
- `MongoCollection.bulkUpsert` (and Postgres / FlatPostgres) returned `true` whenever `bulkWrite` / `executeBatch` did not throw, without checking whether every requested document was actually written.
- That loose contract let callers such as attribute-service treat incomplete upserts as success (2xx), which is hard to detect upstream.
- Now validate write completeness before returning success:
  - **Mongo:** `wasAcknowledged` and `matchedCount + upserts.size() == requested`
  - **Postgres / FlatPostgres:** JDBC batch length matches submitted ops and no `EXECUTE_FAILED` (via `PostgresCollection.isBatchFullySuccessful`)
- `bulkUpsertAndReturnOlderDocuments` throws `IOException` on incomplete writes (Mongo + Postgres JSONB).
- Hard failures (exceptions) still map to `false` / `IOException` as before.

## Context
Discovered while investigating attribute-service create returning success when attributes were missing after wipe/recreate. Attribute-service adds a post-write verify as defense in depth; this fixes the shared document-store contract for all consumers.

## Test plan
- [x] Unit: `MongoCollectionTest.BulkUpsert` (complete / incomplete / unacknowledged)
- [x] Unit: `PostgresCollectionTest.BatchFullySuccessful`
- [ ] CI integration tests on PR


document-store/src/main/java/org/hypertrace/core/documentstore/commons/BatchWriteUtils.java
```diff
@@ -0,0 +1,27 @@
...
+
+  private BatchWriteUtils() {}
+
+  /**
+   * Returns true when every batch entry completed without {@link Statement#EXECUTE_FAILED} and the
+   * result length matches the number of operations submitted.
+   *
+   * <p>{@link Statement#SUCCESS_NO_INFO} (-2) and positive update counts are treated as success.
+   */
+  public static boolean isBatchFullySuccessful(final int[] updateCounts, final int expectedSize) {
```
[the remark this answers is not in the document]
Bhuvan506: Done — moved `isBatchFullySuccessful` onto `PostgresCollection` (used from there and `FlatPostgresCollection`) and dropped `BatchWriteUtils`.

saxenakshitiz (approved):

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/PostgresCollectionTest.java
```diff
@@ -1237,6 +1239,38 @@ void testUpsertSQLException() throws SQLException, IOException {
     verify(mockUpsertPreparedStatement, times(1)).setString(eq(3), any());
   }
 
+  @Nested
+  class BatchFullySuccessful {
+
+    @Test
+    void allPositive_returnsTrue() {
```
suddendust: Lets stick to camelCase for consistency.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
```diff
@@ -714,6 +741,17 @@ public CloseableIterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Do
...
+      if (!isBulkUpsertComplete(result, documents.size())) {
+        LOGGER.error(
+            "Incomplete bulk upsert for documents. requested={}, matched={}, upserted={},"
+                + " acknowledged={}, result={}",
+            documents.size(),
+            result.wasAcknowledged() ? result.getMatchedCount() : -1,
+            result.wasAcknowledged() ? result.getUpserts().size() : -1,
+            result.wasAcknowledged(),
+            result);
+        throw new IOException("Incomplete bulk upsert.");
```
suddendust: This might now break existing clients.

suddendust: We should close the cursor here or it'll be a resource leak.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -801,6 +809,13 @@ public CloseableIterator<Document> bulkUpsertAndReturnOlderDocuments(Map<Key, Do
       if (LOGGER.isDebugEnabled()) {
         LOGGER.debug("Write result: {}", Arrays.toString(updateCounts));
       }
+      if (!isBatchFullySuccessful(updateCounts, documents.size())) {
+        LOGGER.error(
+            "Incomplete bulk upsert for documents. requested={}, updateCounts={}",
+            documents.size(),
+            Arrays.toString(updateCounts));
+        throw new IOException("Incomplete bulk upsert.");
```
suddendust: Close the RS here to prevent leak?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
```diff
@@ -1037,6 +1052,25 @@ private int[] bulkUpsertImpl(Map<Key, Document> documents) throws SQLException,
...
 
+  /**
+   * Returns true when every batch entry completed without {@link Statement#EXECUTE_FAILED} and the
+   * result length matches the number of operations submitted.
+   *
+   * <p>{@link Statement#SUCCESS_NO_INFO} (-2) and positive update counts are treated as success.
+   */
+  @VisibleForTesting
+  static boolean isBatchFullySuccessful(final int[] updateCounts, final int expectedSize) {
+    if (updateCounts == null || updateCounts.length != expectedSize) {
```
suddendust: This method will always return `true` due to the way we're building the batch. Because we add 1 doc per batch, `updateCounts.length == expectedSize` will always hold true. Also, if a single doc/batch failed upsert, it'll throw `BatchUpdateException`. 

Rather, smth like this should help:
```
} catch (BatchUpdateException e) {
  int[] partial = e.getUpdateCounts();
  LOGGER.error(
      "BatchUpdateException bulk inserting documents. requested={}, updateCounts={}",
      documents.size(),
      Arrays.toString(partial),
      e);
  return false;   // partial application: some entries succeeded, some EXECUTE_FAILED
}
```

Can you check this?

suddendust:
Lets add some integration tests here to validate the behaviour consistency b/w Mongo and PG.

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/PostgresCollectionTest.java
Bhuvan506: Removed those nested tests (Postgres success-path batch-length check was dropped per your later comment). New IT method names are camelCase.

document-store/src/main/java/org/hypertrace/core/documentstore/mongo/MongoCollection.java
Bhuvan506: Agreed it tightens the contract: callers that previously treated incomplete Mongo writes as success will now get `false` / `IOException`. That is intentional for the gap we care about. Attribute-service is also adding a client-side post-write verify + evidence logs; we can stage rollout if needed.

Bhuvan506: Fixed — close the Mongo cursor before throwing on incomplete upsert (and on other failure paths before the iterator takes ownership).

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
Bhuvan506: Fixed — close the ResultSet in a finally when we do not hand it to `PostgresResultIterator`.

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresCollection.java
Bhuvan506: Checked — agree. Dropped the success-path `isBatchFullySuccessful` check. `BatchUpdateException` now logs `requested` + `e.getUpdateCounts()` and returns false / fails `bulkUpsertAndReturnOlderDocuments`.

Bhuvan506:
Added Mongo/PG consistency ITs in `MongoPostgresWriteConsistencyTest.BulkUpsertConsistencyTest`: complete-batch success + readability, and `bulkUpsertAndReturnOlderDocuments` behaviour on both stores.

PR #323, opened 2026-08-27
Add ALL and ONE array filter operators for MongoDB and Postgres

lrathod:
## Summary

Adds two new `ArrayOperator` values for filtering on array-valued attributes (ticket ASP-3008):

- **ALL** — array attribute must contain *every* value specified in the filter
- **ONE** — array attribute must contain *exactly one* element, and that element must be one of the specified values

### MongoDB
- `ALL` → `{"$expr": {"$setIsSubset": [<values>, {"$ifNull": ["$<arrayPath>", []]}]}}` (set containment, null-safe)
- `ONE` → `$and` of `$size == 1` and `$in` on `$arrayElemAt [path, 0]` (cardinality + membership)

### Postgres
- **Native array columns** (flat collections): `ALL` → `COALESCE(col, ARRAY[]::type[]) @> ?`; `ONE` → `array_length(...) = 1 AND col && ?`
- **JSONB array paths** (nested documents): `ALL` → `(CASE WHEN jsonb_typeof(path) = 'array' THEN path ELSE '[]'::jsonb END) @> ?::jsonb`; `ONE` → `jsonb_array_length(...) = 1 AND (path @> ?::jsonb OR ...)` per value
- SQL array type inferred from the constant values via `PostgresDataType.fromJavaValue()`

### Design note
Both parsers require the inner `RelationalExpression` to carry a **constant value list**; a non-constant RHS throws `UnsupportedOperationException`. This is intentional — `ALL`/`ONE` are set-level operators, unlike `ANY` which supports arbitrary per-element sub-filters.

## Test plan

- [x] 6 new unit tests in `MongoArrayFilterParserTest` (operator structure, `$ifNull` guards, single-value, non-constant RHS rejection, no double `$expr` wrapping)
- [x] 5 new unit tests in `PostgresQueryParserTest` (ALL/ONE × JSONB/native array, nested JSONB path)
- [x] 2 new parameterized integration tests in `ArrayFiltersQueryIntegrationTest` (Mongo + Postgres) with 3 new JSON fixtures — run in CI
- [x] `:document-store:build` (compile + unit tests + spotless) passes locally

Made with [Cursor](https://cursor.com)

### DIFF
--- a/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/ArrayFiltersQueryIntegrationTest.java
+++ b/document-store/src/integrationTest/java/org/hypertrace/core/documentstore/ArrayFiltersQueryIntegrationTest.java
@@ -2,7 +2,9 @@ package org.hypertrace.core.documentstore;
 
 import static org.hypertrace.core.documentstore.expression.impl.LogicalExpression.and;
 import static org.hypertrace.core.documentstore.expression.impl.LogicalExpression.not;
+import static org.hypertrace.core.documentstore.expression.operators.ArrayOperator.ALL;
 import static org.hypertrace.core.documentstore.expression.operators.ArrayOperator.ANY;
+import static org.hypertrace.core.documentstore.expression.operators.ArrayOperator.ONE;
 import static org.hypertrace.core.documentstore.model.config.DatabaseType.MONGO;
 import static org.hypertrace.core.documentstore.model.config.DatabaseType.POSTGRES;
 import static org.hypertrace.core.documentstore.utils.Utils.MONGO_STORE;
@@ -11,6 +13,7 @@ import static org.hypertrace.core.documentstore.utils.Utils.POSTGRES_STORE;
 import com.google.common.io.Resources;
 import java.io.IOException;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Spliterator;
 import java.util.Spliterators;
@@ -337,6 +340,84 @@ class ArrayFiltersQueryIntegrationTest {
 
     final Iterator<Document> documents = collection.aggregate(query);
     final String expected = readResource("environment_ids_subset.json");
+    final String actual = iteratorToJson(documents);
+
+    datastore.deleteCollection(testCollectionName);
+
+    JSONAssert.assertEquals(expected, actual, JSONCompareMode.LENIENT);
+  }
+
+  /**
+   * Tests MATCH_ALL semantics: documents whose array attribute contains every value specified in
+   * the filter.
+   */
+  @ParameterizedTest
+  @ArgumentsSource(AllProvider.class)
+  void getDocumentsContainingAllGivenValues(final String dataStoreName)
+      throws JSONException, IOException {
+    final String testCollectionName = "array_match_test";
+    final Datastore datastore = datastoreMap.get(dataStoreName);
+    final Map<Key, Document> testDocuments =
+        Utils.buildDocumentsFromResource("query/array_operators/array_match_test.json");
+    datastore.deleteCollection(testCollectionName);
+    datastore.createCollection(testCollectionName, null);
+    final Collection collection = datastore.getCollection(testCollectionName);
+    collection.bulkUpsert(testDocuments);
+
+    final Query query =
+        Query.builder()
+            .setFilter(
+                ArrayRelationalFilterExpression.builder()
+                    .operator(ALL)
+                    .filter(
+                        RelationalExpression.of(
+                            IdentifierExpression.of("tags"),
+                            RelationalOperator.IN,
+                            ConstantExpression.ofStrings(List.of("red", "blue"))))
+                    .build())
+            .build();
+
+    final Iterator<Document> documents = collection.aggregate(query);
+    final String expected = readResource("array_match_all_result.json");
+    final String actual = iteratorToJson(documents);
+
+    datastore.deleteCollection(testCollectionName);
+
+    JSONAssert.assertEquals(expected, actual, JSONCompareMode.LENIENT);
+  }
+
+  /**
+   * Tests MATCH_ONE semantics: documents whose array attribute has exactly one element, and that
+   * element is one of the values specified in the filter.
+   */
+  @ParameterizedTest
+  @ArgumentsSource(AllProvider.class)
+  void getDocumentsWithExactlyOneElementMatchingGivenValues(final String dataStoreName)
+      throws JSONException, IOException {
+    final String testCollectionName = "array_match_test";
+    final Datastore datastore = datastoreMap.get(dataStoreName);
+    final Map<Key, Document> testDocuments =
+        Utils.buildDocumentsFromResource("query/array_operators/array_match_test.json");
+    datastore.deleteCollection(testCollectionName);
+    datastore.createCollection(testCollectionName, null);
+    final Collection collection = datastore.getCollection(testCollectionName);
+    collection.bulkUpsert(testDocuments);
+
+    final Query query =
+        Query.builder()
+            .setFilter(
+                ArrayRelationalFilterExpression.builder()
+                    .operator(ONE)
+                    .filter(
+                        RelationalExpression.of(
+                            IdentifierExpression.of("tags"),
+                            RelationalOperator.IN,
+                            ConstantExpression.ofStrings(List.of("red", "blue"))))
+                    .build())
+            .build();
+
+    final Iterator<Document> documents = collection.aggregate(query);
+    final String expected = readResource("array_match_one_result.json");
     final String actual = iteratorToJson(documents);
 
     datastore.deleteCollection(testCollectionName);
--- a/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
+++ b/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
@@ -8,6 +8,9 @@ import static org.hypertrace.core.documentstore.postgres.PostgresCollection.ID;
 import static org.hypertrace.core.documentstore.postgres.utils.PostgresUtils.encodeAliasForNestedField;
 import static org.hypertrace.core.documentstore.postgres.utils.PostgresUtils.prepareParsedNonCompositeFilter;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
@@ -15,6 +18,7 @@ import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
 import org.hypertrace.core.documentstore.DocumentType;
 import org.hypertrace.core.documentstore.Key;
+import org.hypertrace.core.documentstore.expression.impl.ArrayFilterExpression;
 import org.hypertrace.core.documentstore.expression.impl.ArrayRelationalFilterExpression;
 import org.hypertrace.core.documentstore.expression.impl.ConstantExpression;
 import org.hypertrace.core.documentstore.expression.impl.DocumentArrayFilterExpression;
@@ -24,6 +28,7 @@ import org.hypertrace.core.documentstore.expression.impl.LogicalExpression;
 import org.hypertrace.core.documentstore.expression.impl.RelationalExpression;
 import org.hypertrace.core.documentstore.expression.operators.LogicalOperator;
 import org.hypertrace.core.documentstore.expression.type.FilterTypeExpression;
+import org.hypertrace.core.documentstore.expression.type.SelectTypeExpression;
 import org.hypertrace.core.documentstore.parser.FilterTypeExpressionVisitor;
 import org.hypertrace.core.documentstore.postgres.query.v1.PostgresQueryParser;
 import org.hypertrace.core.documentstore.postgres.query.v1.parser.builder.PostgresSelectExpressionParserBuilder;
@@ -117,7 +122,10 @@ public class PostgresFilterTypeExpressionVisitor implements FilterTypeExpression
     switch (expression.getOperator()) {
       case ANY:
         return getFilterStringForAnyOperator(expression);
-
+      case ALL:
+        return getFilterStringForAllOperator(expression);
+      case ONE:
+        return getFilterStringForOneOperator(expression);
       default:
         throw new UnsupportedOperationException(
             "Unsupported array operator: " + expression.getOperator());
@@ -137,7 +145,10 @@ public class PostgresFilterTypeExpressionVisitor implements FilterTypeExpression
     switch (expression.getOperator()) {
       case ANY:
         return getFilterStringForAnyOperator(expression);
-
+      case ALL:
+        return getFilterStringForAllOperator(expression);
+      case ONE:
+        return getFilterStringForOneOperator(expression);
       default:
         throw new UnsupportedOperationException(
             "Unsupported array operator: " + expression.getOperator());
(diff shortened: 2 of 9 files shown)


document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -332,4 +343,157 @@ private String getFilterStringForAnyOperator(final DocumentArrayFilterExpression
...
+              + expression.getOperator()
+              + " requires a constant list of values, got: "
+              + rhs);
+    }
+
+    final Object value = ((ConstantExpression) rhs).getValue();
+    return value instanceof List ? (List<?>) value : List.of(value);
+  }
+
+  private PostgresDataType resolvePostgresDataType(final List<?> values) {
```
suddendust: Can we avoid a runtime check? `ArrayIdentifierExpression` contains `DataType` that can be extracted statically.

suddendust:
What about operators for nested json arrays?

suddendust:
Lets add some integration tests? You can add in `DocStoreQueryV1Test`.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/ArrayFiltersQueryIntegrationTest.java
```diff
@@ -344,6 +347,84 @@ void getDocumentsWithEnvironmentIdsSubsetOfGivenList(final String dataStoreName)
     JSONAssert.assertEquals(expected, actual, JSONCompareMode.LENIENT);
   }
 
+  /**
+   * Tests MATCH_ALL semantics: documents whose array attribute contains every value specified in
+   * the filter.
+   */
+  @ParameterizedTest
+  @ArgumentsSource(AllProvider.class)
+  void getDocumentsContainingAllGivenValues(final String dataStoreName)
```
suddendust: Oh I see we already have ITs. Can we move them to `DocStoreQueryV1Test`?

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
lrathod: Done. The native-array branch now resolves the element type from the compile-time type info on the field expression (`ArrayIdentifierExpression#getElementDataType()` / `IdentifierExpression#getDataType()`) via `PostgresDataType.fromDataType`, and only falls back to inferring from the filter values when the field carries no type info (UNSPECIFIED). The runtime `jsonb_typeof` guard is now retained only for the JSONB/nested-array path, where the value is schemaless and can be JSON null or a non-array at runtime. Added unit tests covering both the compile-time precedence (declared long[] wins over Integer values) and the fallback.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/ArrayFiltersQueryIntegrationTest.java
lrathod: sure, let me move

lrathod: Done. Moved all ALL/ONE integration tests into `DocStoreQueryV1Test` as a nested `ArrayMatchAllOneOperatorTest` class and removed them from `ArrayFiltersQueryIntegrationTest`. They now run against the shared collections: nested JSONB path (`props.colors`) on both stores, native array columns (`tags`/`flags`) on the flat collection using the existing `PostgresArrayTypeProvider` (typed + untyped variants), and the JSONB array column on the flat collection.

lrathod:
Addressed the review feedback in the latest commits:

**Nested JSON arrays** — ALL/ONE already go through the same array-source resolution infra as ANY (`PostgresFieldIdentifierExpressionVisitor` / `MongoDollarPrefixingIdempotentParser`), so nested paths like `props.colors` and `scope.environmentScope.environmentIds` work. Added coverage to prove it: unit tests in `PostgresQueryParserTest` (nested ALL + ONE) and `MongoArrayFilterParserTest` (nested ALL + ONE), plus integration tests in `DocStoreQueryV1Test` running nested-path ALL/ONE against both MongoDB and Postgres.

**Integration tests** — moved to `DocStoreQueryV1Test` (nested `ArrayMatchAllOneOperatorTest`), covering: nested JSONB array field (both stores), native array columns on the flat collection with typed and untyped `ArrayIdentifierExpression` (via `PostgresArrayTypeProvider`), and the JSONB array column on the flat collection.

**Duplicates question** — `[red, red] ALL [red]` returns **true** in both backends. `` treats both operands as sets (duplicates collapsed), and Postgres `@>` is element-wise containment (each RHS element must exist in LHS; duplicates on either side are irrelevant). So ALL follows set semantics — order and duplicates don't matter. Added an integration test (`testAllWithDuplicatesInDocumentArray`) that asserts a document with `tags: [red, red]` matches `ALL [red]` on both datastores. (For completeness: ONE is unaffected by this — `[red, red]` has length 2, so it never matches ONE.)

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
lrathod: sure, let me update, 
for nested arrays, we'd still need runtime check

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,383 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+                      .operator(ArrayOperator.EXACTLY_ONE)
+                      .filter(
+                          RelationalExpression.of(
+                              IdentifierExpression.of("props.colors"),
+                              IN,
+                              ConstantExpression.ofStrings(List.of("Black", "White"))))
+                      .build())
+              .build();
+      // Only id 3 has exactly one element, and it is Black
+      assertEquals(1, collection.count(oneBlackOrWhite));
```
suddendust: Nit: Would be better to assert on the actual docIds for 100% confidence.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,383 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+                              ConstantExpression.ofStrings(List.of("Black", "White"))))
+                      .build())
+              .build();
+
+      // Only id 3 has exactly one element, and it is Black
+      assertEquals(1, flatCollection.count(query));
+    }
+
+    /**
+     * A non-array value (here props.brand, a string) must simply not match instead of failing the
```
suddendust: If this how we're handling other operators too? Can you validate what happens when the LHS type does not conform to the RHS for an existing operator?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,383 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+     * query - Postgres uses the jsonb_typeof guard, MongoDB an $isArray guard.
+     */
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testAllAndOneOnNonArrayValueDoNotMatch(String dataStoreName) {
```
suddendust: Also, what happens in this case for top-level non-array fields?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,383 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+      collection.bulkUpsert(testDocuments);
+
+      Query allQuery =
+          Query.builder()
+              .setFilter(
+                  ArrayRelationalFilterExpression.builder()
+                      .operator(ArrayOperator.ALL)
+                      .filter(
+                          RelationalExpression.of(
+                              IdentifierExpression.of("props.metadata.colors"),
```
suddendust: Should we not use `JsonIdentifierExpression` here?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,383 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+
+      Query allBlueAndGreen =
+          Query.builder()
+              .setFilter(
+                  ArrayRelationalFilterExpression.builder()
+                      .operator(ArrayOperator.ALL)
+                      .filter(
+                          RelationalExpression.of(
+                              IdentifierExpression.of("props.colors"),
+                              IN,
```
suddendust: What does `IN` mean here semantically?

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
```diff
@@ -1945,4 +1948,335 @@ void testNotInOperatorWithScalarStringField() {
     assertEquals("text", arrayParam.getSqlType());
     assertEquals(2, arrayParam.getValues().length);
   }
+
+  @Test
+  void testAllOperatorWithJsonbArrayField() {
```
suddendust: So this is the legacy PG flow, with no support for first-class columns. That's why you'll see everything is accessed via `->` accessor (for ex: `jsonb_typeof(document->'tags')`). What we need is the flat collection flow. You can parse queries for flat collections using:

```
    PostgresQueryParser postgresQueryParser =
        new PostgresQueryParser(TEST_TABLE, PostgresQueryTransformer.transform(query), new FlatPostgresFieldTransformer());
```

Can you please add test cases for that? I am actually surprised that I am unable to find a test class that tests parsed queries for flat collections. If that is indeed the case, would you mind creating one? Thanks :)

suddendust: But do keep these tests

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: for Postgres flat native column (item as text) scalar type non-array throws we say  "operator does not exist: text @> text[]" 


Document JSONB (mongo + postgres item) ->  “top-level non-array → no match,”

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: IN is element membership, not a second array-IN.

ALL + IN [a,b]: every RHS value is in the stored array (extras on the array OK).
EXACTLY_ONE + IN [a,b]: stored array length is 1, and that element is in {a,b}.
So EXACTLY_ONE ["Code","Live Traffic"] = singleton Code or singleton Live Traffic, not the two-element array ["Code","Live Traffic"]. 

Added test and also add the same in PR desc as well

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: JsonIdentifierExpression on nested document collections — NestedPostgresColTransformer rejects it. Nested ITs still use IdentifierExpression.of("props.colors"). Flat JSONB does use JsonIdentifierExpression.of("props", "colors"). Honest reply: we used it where the transformer supports it; document nested path still uses dotted identifiers.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: We validated ALL/EXACTLY_ONE type mismatch against existing EQ and IN in this suite (string vs number).

- Mongo EQ and IN: 0 rows, no throw.
- Postgres EQ of `"10"` on numeric `quantity`: matches (JSON `->>` is text, so `"10"` equals stored 10).
- Postgres IN of `["10"]` on the same field: 0 rows (jsonb containment is typed).

ALL/EXACTLY_ONE with numeric RHS against string array `props.colors`: 0 rows on both Mongo and Postgres — same family as Mongo EQ/IN and Postgres IN, not Postgres EQ's text coercion. No new error path; mismatch is empty result for this JSON/document case.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,397 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+    }
+
+    /**
+     * A non-array value (here props.brand, a string) must simply not match instead of failing the
+     * query - Postgres uses the jsonb_typeof guard, MongoDB an $isArray guard.
+     */
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testAllAndOneOnNonArrayValueDoNotMatch(String dataStoreName) {
+      Collection collection = getCollection(dataStoreName);
```
suddendust: Let us run these tests on flat collections please: `Collection flatCollection = getFlatCollection(dataStoreName);`. 

`Collection collection = getCollection(dataStoreName);`: This returns the legacy storage mode PG in and therefore it's using `jsonb_typeof` guard.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,397 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+      datastore.deleteCollection(testCollectionName);
+    }
+
+    /**
+     * Documents that ALL is order-independent: ["red", "blue"] ALL ["blue", "red"] is true in both
+     * backends ($setIsSubset and @> are both set containment).
+     */
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testAllIsOrderIndependent(String dataStoreName) throws IOException {
```
suddendust: Should we also validate this for flat?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,397 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+    }
+
+    /**
+     * Documents the set-containment semantics of ALL: duplicates in the document's array do not
+     * affect the outcome. ["red", "red"] ALL ["red"] is true both in MongoDB ($setIsSubset treats
+     * its operands as sets) and in Postgres (@> is element-wise containment).
+     */
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testAllWithDuplicatesInDocumentArray(String dataStoreName) throws IOException {
```
suddendust: Same as above, validate for flat?

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
```diff
@@ -7141,6 +7141,397 @@ void testDateConstantExpressionInFilterForPostgresThrowsException(String dataSto
...
+      datastore.deleteCollection(testCollectionName);
+    }
+
+    /**
+     * Documents that EXACTLY_ONE counts raw elements, not distinct values: a document with tags
+     * ["red", "red"] does NOT match EXACTLY_ONE ["red"] because the array has two elements.
+     */
+    @ParameterizedTest
+    @ArgumentsSource(AllProvider.class)
+    void testOneWithDuplicatesInDocumentArray(String dataStoreName) throws IOException {
```
suddendust: Same ^

suddendust:
Can you add a test for parsed queries for flat collections? We might need that information for perf tuning.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: Done. Added `testAllAndOneOnNonArrayValueDoNotMatchFlat` — same ALL/EXACTLY_ONE no-match assertion on the flat collection via `JsonIdentifierExpression.of("props", "brand")` (flat JSONB scalar, exercises the flat path instead of the legacy `jsonb_typeof` document guard).

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: Done. Added `testAllIsOrderIndependentFlat` — same fixture as `array_match_test.json` loaded into a temp flat table with a native `tags TEXT[]` column, run with both WITH_TYPE and WITHOUT_TYPE (`PostgresArrayTypeProvider`). ALL ["blue","red"] matches ids {1,2} on the flat native-array path too.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: Done. Added `testAllWithDuplicatesInDocumentArrayFlat` (temp flat table, native `tags TEXT[]`, WITH_TYPE/WITHOUT_TYPE). ALL ["red"] matches ids {1,2,3,7} — `{red,red}` matches ALL ["red"] on the flat path as well (@> is element-wise containment).

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: Done. Added `testOneWithDuplicatesInDocumentArrayFlat` (temp flat table, native `tags TEXT[]`, WITH_TYPE/WITHOUT_TYPE). EXACTLY_ONE ["red"] matches only id 3 — `{red,red}` is excluded because array_length = 2, same as the document path.

lrathod:
Yes — parsed-query tests for flat collections are in `PostgresQueryParserTest` (all use `FlatPostgresFieldTransformer`). The final SQL shapes:

**Flat, native array column** (`testAllOperatorWithNativeArrayField`, `testOneOperatorWithNativeArrayField`, + BOOLEAN[] / long[] variants):
```sql
-- ALL
SELECT * FROM "testCollection" WHERE "tags" @> ?                      -- param: text[]
-- EXACTLY_ONE
SELECT * FROM "testCollection" WHERE array_length("tags", 1) = 1 AND "tags" && ?
```

**Flat, JSONB column** (`testAllOperatorWithJsonIdentifierOnFlatJsonbColumn`, `testOneOperatorWithJsonIdentifierOnFlatJsonbColumn`, + multi-level path variants):
```sql
-- ALL
... WHERE (CASE WHEN jsonb_typeof("props"->'colors') = 'array'
           THEN "props"->'colors' ELSE '[]'::jsonb END) @> ?::jsonb
-- EXACTLY_ONE
... WHERE jsonb_array_length((CASE ... END)) = 1 AND (CASE ... END) <@ ?::jsonb
```

Index notes:
- Native arrays: both `@>` and `&&` are GIN-indexable (`USING GIN (tags)`); `array_length(...) = 1` is a cheap residual.
- Flat JSONB: the `CASE WHEN jsonb_typeof(...)` guard wraps the expression, so a plain GIN index on the `"props"` column will NOT be used — it would need an expression index on the full CASE expression, or we drop/avoid the guard where the schema guarantees arrays.

Also covered: compile-time element type (from `ArrayIdentifierExpression`) preferred over value-based inference, non-constant RHS rejected, empty RHS rejected at construction.

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
```diff
@@ -1945,4 +1948,512 @@ void testNotInOperatorWithScalarStringField() {
...
+            .build();
+
+    PostgresQueryParser postgresQueryParser =
+        new PostgresQueryParser(
+            TEST_TABLE,
+            PostgresQueryTransformer.transform(query),
+            new FlatPostgresFieldTransformer());
+
+    String sql = postgresQueryParser.parse();
+    assertEquals(
```
suddendust:
```
WHERE jsonb_typeof("props"->'colors') = 'array'
  AND jsonb_array_length("props"->'colors') = 1
  AND "props"->'colors' <@ ?::jsonb
```
Is this better?

lrathod: Good point — containment (@> and <@) is total on non-array jsonb: it returns false on scalars and NULL on missing keys, never errors. So we've dropped the CASE guard for ALL entirely (`WHERE "props"->'colors' @> ?::jsonb`) and for the EXACTLY_ONE `<@` conjunct. The one place a guard is still required is jsonb_array_length — it raises `cannot get array length of a scalar` on non-arrays, and since Postgres doesn't guarantee WHERE conjunct evaluation order, an AND-chained jsonb_typeof check isn't a safe guard; the length check keeps a CASE: `(CASE WHEN jsonb_typeof(...) = 'array' THEN jsonb_array_length(...) ELSE 0 END) = 1 AND "props"->'colors' <@ ?::jsonb`. Unit + integration tests updated.

document-store/src/integrationTest/java/org/hypertrace/core/documentstore/DocStoreQueryV1Test.java
lrathod: Done. All ALL/EXACTLY_ONE integration tests in `ArrayMatchAllOneOperatorTest` now assert exact docId sets via `collectNormalizedIds` (e.g. `assertEquals(Set.of("1", "2", "3", "7"), ...)`), with fixture comments noting which ids match and why. This also implicitly pins the NULL/empty-array exclusions, and there is now an explicit `testAllAndOneExcludeNullAndEmptyArrays` for that guarantee.

document-store/src/test/java/org/hypertrace/core/documentstore/postgres/query/v1/PostgresQueryParserTest.java
lrathod: Done. Added flat-collection parser tests using `PostgresQueryTransformer.transform(query)` + `FlatPostgresFieldTransformer`, and kept the legacy-flow tests as you asked. Flat coverage now includes: native array columns (`testAllOperatorWithNativeArrayField` / `testOneOperatorWithNativeArrayField`, BOOLEAN[] variants, `"tags" @> ?` / `array_length("tags", 1) = 1 AND "tags" && ?`), flat JSONB columns via `JsonIdentifierExpression` (single- and multi-level paths), and compile-time field type vs value-based inference (`testAllOperatorPrefersCompileTimeFieldTypeOverValueInference` / fallback test).

document-store/src/main/java/org/hypertrace/core/documentstore/postgres/query/v1/vistors/PostgresFilterTypeExpressionVisitor.java
```diff
@@ -332,4 +346,183 @@ private String getFilterStringForAnyOperator(final DocumentArrayFilterExpression
...
+          resolvePostgresDataType(expression.getArraySource(), values);
+      postgresQueryParser.getParamsBuilder().addArrayParam(values.toArray(), dataType.getSqlType());
+      return String.format(
+          "array_length(%s, 1) = 1 AND %s && ?",
+          fieldContext.parsedLhs(), fieldContext.parsedLhs());
+    }
+
+    final String guardedLength =
+        String.format(
+            "(CASE WHEN jsonb_typeof(%s) = 'array' THEN jsonb_array_length(%s) ELSE 0 END)",
```
suddendust: We still need case @lrathod ?

lrathod: only for EXACTLY_ONE length check on JSONB fields 

jsonb_array_length() raises cannot get array length of a scalar on non-arrays. And we can't just AND-chain jsonb_typeof(x) = 'array' AND jsonb_array_length(x) = 1 because Postgres doesn't guarantee WHERE conjunct evaluation order — the planner may evaluate the length call first. CASE is the only order-forcing construct → kept, but only around the length check:

